PR #1, opened 2016-09-29
Don't show card payment option if not configured

bluk:
Don't show the credit card payment method if only configured for using other payment methods (e.g. PayPal / Venmo only).


BraintreeDropIn/BTPaymentSelectionViewController.m
```diff
@@ -187,9 +187,16 @@ - (void)configurationLoaded:(__unused BTConfiguration *)configuration error:(NSE
                 }
             }
 
-            // Always add Cards
-            [activePaymentOptions addObject:@(BTUIKPaymentOptionTypeUnknown)];
-            
+            NSArray *supportedCardTypes = [self.configuration.json[@"creditCards"][@"supportedCardTypes"] asArray];
+            for (NSString *supportedCardType in supportedCardTypes) {
+                BTUIKPaymentOptionType paymentOptionType = [BTUIKViewUtil paymentOptionTypeForPaymentInfoType:supportedCardType];
+                if ([BTUIKViewUtil isPaymentOptionTypeACreditCard:paymentOptionType]) {
```
sdcoffey: Is there any other place we would use this new method? Or can we simplify this logic to:

```objc
if ([supportedCardTypes count] > 0) {
     [activePaymentOptions addObject:@(BTUIKPaymentOptionTypeUnknown)];
}
```

sdcoffey: I guess a better question is, is there anything that we would get back in that `supportedCardTypes` array that's not a credit card?

bluk: The supported card types array returned can include Apple Pay/Android Pay "card" information which is why we need to do a slightly deeper inspection.

sdcoffey: Ah, got it. I want to say those card types are also returned in the specific configuration blocks for A*Pay payment methods, so that might be something we can optimize in the future.

Looks good!

PR #7, opened 2016-11-18
Add a public method to access visual assets.

demerino:
See https://github.com/braintree/braintree_ios/issues/289

Rather than exposing the headers - I've opted to follow our existing pattern where assets are fetched via `BTUIKViewUtils`.


bluk:
👍

Issue #8, opened 2016-11-19
typo in BTUIKLocalizedString


miche-atucha:
methods "insertIntoLozalizedString", should be "insertIntoLocalizedString"

demerino:
Thanks @miche-atucha 

I've created PR for the fix which we can release next week: https://github.com/braintree/braintree-ios-drop-in/issues/8

Let us know if you have any other suggestions or comments!

PR #9, opened 2016-11-19
Rename insertIntoLozalizedString to insertIntoLocalizedString.

demerino:
See https://github.com/braintree/braintree-ios-drop-in/issues/8

This is a breaking change - but considering the undocumented/recent introduction of the typo and the fact that we are still marking the SDK as `beta` - I think we can treat this as a bug.


lkorth:
👍

PR #11, opened 2016-12-26
Include of non-modular header inside framework module error in BTPaymentSelectionViewController

steveholt55:
This fixes an issue I was occasionally seeing when trying to add in the DropIn UI and Braintree/ApplePay into my project using Cocoapods.


demerino:
Thanks for submitting the PR, we'll review it after the Holidays.

PR #16, opened 2017-01-13
Add support for card.io

demerino:
Checks to see if card.io classes are available (rather than requiring it explicitly). If it is available a button is added above the keyboard when entering the card number that will start card.io to collect the card number.

To try it in your own project you must:
```
//Update your info.plist to include NSCameraUsageDescription
<key>NSCameraUsageDescription</key>
<string>Entering credit card info via the camera</string>
```
```
//Update your podfile with card.io
pod 'CardIO'
```
```
// Test on a real device with a camera
```
Still some TODOs (like version checking card.io) and tests that need to be added.


steveholt55:
@demerino Just gave this a run. It works great, thanks for the quick action!

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -274,6 +286,55 @@ - (void)updateRequiredFields {
...
+    Class kCardIOUtilities = NSClassFromString(@"CardIOUtilities");
+    if (kCardIOView != nil && kCardIOView != nil
+        && [kCardIOUtilities respondsToSelector:@selector(libraryVersion)]
+        && [kCardIOUtilities respondsToSelector:@selector(canReadCardWithCamera)]) {
+        NSString *cardIOVersion = [kCardIOUtilities libraryVersion];
+        NSString *majorVersion = [cardIOVersion length] >= 2 ? [cardIOVersion substringToIndex:2] : @"";
+        // Require 5.x.x strictly
+        return [majorVersion isEqualToString:@"5."] && [kCardIOUtilities canReadCardWithCamera];
+    }
+    return false;
```
bluk: Should probably be `NO` here.

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -274,6 +286,55 @@ - (void)updateRequiredFields {
...
+    if ([self isCardIOAvailable]) {
+        BTUIKInputAccessoryToolbar *cardToolBar = [BTUIKInputAccessoryToolbar new];
+        UIBarButtonItem *flexSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFlexibleSpace target:nil action:nil];        
+        UIBarButtonItem *cardIOButton = [[BTUIKBarButtonItem alloc] initWithTitle:BTUIKLocalizedString(SCAN_CARD_IO_ACTION) style:UIBarButtonItemStylePlain target:self action:@selector(presentCardIO)];
+        cardToolBar.items = @[flexSpace, cardIOButton, flexSpace];
+        self.cardNumberField.textField.inputAccessoryView = cardToolBar;
+    }
+}
+
+- (BOOL)isCardIOAvailable {
```
bluk: While I think this method would work for people the majority of the time, is there a way to disable it? I don't know who would have `card.io` (maybe as another dependency) and not want it for the card form, but just something to consider. Maybe implement when someone asks?

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -274,6 +286,55 @@ - (void)updateRequiredFields {
...
+    Class kCardIOView = NSClassFromString(@"CardIOPaymentViewController");
+    Class kCardIOUtilities = NSClassFromString(@"CardIOUtilities");
+    if (kCardIOView != nil && kCardIOView != nil
+        && [kCardIOUtilities respondsToSelector:@selector(libraryVersion)]
```
bluk: I would use `cardIOLibraryVersion` especially if this method is called dynamically. https://github.com/card-io/card.io-iOS-source/blob/master/CardIO_Public_API/CardIOUtilities.h#L42 . Apple has flagged a few developers in app review where they used a method called `libraryVersion` thinking they were calling an internal private method.

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -274,6 +286,55 @@ - (void)updateRequiredFields {
...
 
+#pragma mark - card.io
+
+- (void)setupCardIO {
+    if ([self isCardIOAvailable]) {
+        BTUIKInputAccessoryToolbar *cardToolBar = [BTUIKInputAccessoryToolbar new];
+        UIBarButtonItem *flexSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFlexibleSpace target:nil action:nil];        
+        UIBarButtonItem *cardIOButton = [[BTUIKBarButtonItem alloc] initWithTitle:BTUIKLocalizedString(SCAN_CARD_IO_ACTION) style:UIBarButtonItemStylePlain target:self action:@selector(presentCardIO)];
+        cardToolBar.items = @[flexSpace, cardIOButton, flexSpace];
+        self.cardNumberField.textField.inputAccessoryView = cardToolBar;
```
bluk: I would look at using the `preloadCardIO` ( https://github.com/card-io/card.io-iOS-source/blob/master/CardIO_Public_API/CardIOUtilities.h#L49 ) method potentially if more people use card.io to improve the speed. Could be a future enhancement.

demerino: We'll leave it to developers to add this preload for now.

BraintreeDropIn/BTCardFormViewController.m
demerino: Agreed. I decided to not make it opt-in. We can add an option to opt-out when it comes up.

Issue #28, opened 2017-03-15
Add a check to determine whether the customer added a new card or used a recent card


kierobinson1:
In the completion of the Drop-in controller it would be nice to know if the customer used a new card or a recent card. Specifically for us, this would be for use in analytics so we can see how many customers are purchasing with recent cards, but I'm sure there are other uses.

Submitting a pull request shortly showing what this change could look like.

meismyles:
See #29.

demerino:
@kierobinson1 This is a great idea - should we consider adding any other events that could be useful for your analytics?

kierobinson1:
@demerino I can't think of any other events we could track at the minute, just this one!

demerino:
Closing in favor of #29

PR #29, opened 2017-03-15
Add a check to determine whether the customer added a new card or used a vaulted card

kierobinson1:
Would be nice to know if the customer used a new or vaulted card in the completion of the Drop-in for things like analytics etc.

[See issue 28](https://github.com/braintree/braintree-ios-drop-in/issues/28).

### DIFF
--- a/BraintreeDropIn/Public/BTDropInResult.h
+++ b/BraintreeDropIn/Public/BTDropInResult.h
@@ -29,6 +29,9 @@ typedef void (^BTDropInResultFetchHandler)(BTDropInResult * _Nullable result, NS
 /// The payment method nonce
 @property (nonatomic, strong, nullable) BTPaymentMethodNonce *paymentMethod;
 
+/// True if user selected a recent item from the BTPaymentSelectionViewController collectionView.
+@property (nonatomic, assign, getter=isRecentItem) BOOL recentItem;
+
 /// Fetch a BTDropInResult without displaying or initializing a BTDropInController. Works with client tokens that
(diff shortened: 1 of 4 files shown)


demerino:
@kierobinson1 Thanks for submitting this PR - we'll take a look!

Are there other events/analytics that you'd be interested in?

kierobinson1:
@demerino cheers! This is the only event I can think of that'd be useful at the moment.

BraintreeDropIn/Public/BTDropInResult.h
```diff
@@ -29,6 +29,9 @@ typedef void (^BTDropInResultFetchHandler)(BTDropInResult * _Nullable result, NS
 /// The payment method nonce
 @property (nonatomic, strong, nullable) BTPaymentMethodNonce *paymentMethod;
 
+/// True if user selected a vaulted item from the BTPaymentSelectionViewController collectionView.
+@property (nonatomic, assign, getter=isVaultedItem) BOOL vaultedItem;
```
demerino: I wonder if developers could be confused and think this value indicates whether the payment method WAS VAULTED. Maybe we can rename this to something like `selectedFromVault` or `existingPaymentMethod`. Thoughts?

kierobinson1: Yeah thats a good point, I agree!

demerino:
@kierobinson1 Following up - is this still something that you're interested in?

demerino:
Feel free to update this PR if you're still interested.

PR #30, opened 2017-03-15
Handle 3ds cancel event in Drop-in, update Demo app to better document 3DS flow

sdcoffey:
- Returns a `cancel` event when customer exits 3DS validation
- Updates demo with 3DS example


demerino:
Updated this a bit more. No longer returns a cancel even if the customer "cancels" 3DS. Instead they will be returned to the primary view and it will be updated to check for new vaulted payments.

Issue #27, opened 2017-03-15
Detect when 'Add Card' view controller opens so that status bar color can be changed


meismyles:
_**Moved from [braintree_ios issue tracker](https://github.com/braintree/braintree_ios/issues/316).**_

Currently, when the 'Add Card' view controller is presented when using the drop-in UI, there seems to be no way to detect that this has shown, and this means that if using the default light theme, if using a white status bar style, it cannot be seen since it blends in with the nav bar.

I know you also offer a dark theme but it seems like there should either be a way of being notified that the 'Add Card' flow has opened so we can change the status bar style ourselves manually, or in the drop-in framework, maybe in the BTUIKAppearance customisation options, an option should be given to set the status bar style.

demerino:
We'll look into the different options here to see what works best. Do you have a preference?

meismyles:
Personally, I would say the most preferable option is to have the Drop-in handle this itself. You take away work from the developer then. So:
1. Store the current status bar style just before the 'Add Card' view is opened.
2. Depending on the Drop-in theme chosen, apply either the black status bar if light theme chosen, or white status bar if dark theme chosen.
3. Set the status bar style back to stored value on close.

demerino:
See #36 

I do think that Drop-in should do the work. To simplify things, I've created a PR that simply hides the status bar for "full screen" views. Let me know what you think.

demerino:
Fixed in `5.2.0`.

PR #36, opened 2017-03-31
Hide the status bar when showing full screen view controllers

demerino:
Possible fix for https://github.com/braintree/braintree-ios-drop-in/issues/27

Instead of adding more options for the status bar - simply hide it for the full screen modals (card form, union pay enrollment). This is technically a change in default appearance (breaking?). But I'm in favor of trying this before adding more customization. The status bar will still appear as normal when displaying the initial payment selection view.

Note: The status bar is the top bar that contains the time, battery info, etc...

### DIFF
--- a/UnitTests/BTDropInBaseViewControllerTests.m
+++ b/UnitTests/BTDropInBaseViewControllerTests.m
@@ -19,7 +19,12 @@
 
 - (void)test_prefersStatusBarHidden_returnsTrue {
     BTDropInBaseViewController *viewController = [[BTDropInBaseViewController alloc] init];
-    XCTAssertFalse(viewController.prefersStatusBarHidden);
+    XCTAssertTrue(viewController.prefersStatusBarHidden);
+}
+
+- (void)test_preferredStatusBarUpdateAnimation_returns_UIStatusBarAnimationSlide {
+    BTDropInBaseViewController *viewController = [[BTDropInBaseViewController alloc] init];
+    XCTAssertEqual(UIStatusBarAnimationSlide, viewController.preferredStatusBarUpdateAnimation);
 }
 
 @end
--- a/UnitTests/BTDropInBaseViewControllerTests.m
+++ b/UnitTests/BTDropInBaseViewControllerTests.m
@@ -7,22 +7,12 @@
 
 @implementation BTDropInBaseViewControllerTests
 
-- (void)setUp {
-    [super setUp];
-    // Put setup code here. This method is called before the invocation of each test method in the class.
-}
-
-- (void)tearDown {
-    // Put teardown code here. This method is called after the invocation of each test method in the class.
-    [super tearDown];
-}
-
-- (void)test_prefersStatusBarHidden_returnsTrue {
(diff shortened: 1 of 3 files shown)


UnitTests/BTDropInBaseViewControllerTests.m
```diff
@@ -0,0 +1,25 @@
...
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [super tearDown];
+}
+
+- (void)test_prefersStatusBarHidden_returnsTrue {
+    BTDropInBaseViewController *viewController = [[BTDropInBaseViewController alloc] init];
+    XCTAssertFalse(viewController.prefersStatusBarHidden);
```
bluk: Not sure if you also want to add a test for `preferredStatusBarUpdateAnimation`.

demerino: Done.

UnitTests/BTDropInBaseViewControllerTests.m
```diff
@@ -0,0 +1,30 @@
+#import <XCTest/XCTest.h>
+#import "BTDropInBaseViewController.h"
+
+@interface BTDropInBaseViewControllerTests : XCTestCase
+
+@end
+
+@implementation BTDropInBaseViewControllerTests
+
+- (void)setUp {
```
skunkworks: Remove setUp and tearDown

UnitTests/BTDropInBaseViewControllerTests.m
```diff
@@ -0,0 +1,30 @@
...
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [super tearDown];
+}
+
+- (void)test_prefersStatusBarHidden_returnsTrue {
+    BTDropInBaseViewController *viewController = [[BTDropInBaseViewController alloc] init];
+    XCTAssertTrue(viewController.prefersStatusBarHidden);
+}
+
+- (void)test_preferredStatusBarUpdateAnimation_returns_UIStatusBarAnimationSlide {
```
skunkworks: `testPreferredStatusBarUpdateAnimation_returnsUIStatusBarAnimationSlide`

UnitTests/BTDropInBaseViewControllerTests.m
```diff
@@ -0,0 +1,30 @@
...
+    [super setUp];
+    // Put setup code here. This method is called before the invocation of each test method in the class.
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [super tearDown];
+}
+
+- (void)test_prefersStatusBarHidden_returnsTrue {
```
skunkworks: `testPrefersStatusBarHidden_returnsTrue`

meismyles:
I'd say this solution is fine for me, but as @skunkworks says, maybe not for others. Seems reasonable to add this fix in for now as a short term solution rather than it just looking 'broken', and then add styling options in the future.

PR #38, opened 2017-03-31
Update BTUIKCardType to match on regex rather than prefix lists

demerino:
Fixes Maestro `6703` prefix not being found: https://github.com/braintree/braintree_ios/issues/322

Regex was lifted from the Android SDK: https://github.com/braintree/android-card-form/blob/master/CardForm/src/main/java/com/braintreepayments/cardform/utils/CardType.java#L12-L45

### DIFF
--- a/BraintreeUIKit/Models/BTUIKCardType.m
+++ b/BraintreeUIKit/Models/BTUIKCardType.m
@@ -76,21 +76,6 @@
     return nil;
 }
 
-// Since each card type has a list of acceptable card prefixes, we
-// can determine which card types may match a given number
-+ (NSArray *)possibleCardTypesForNumber:(NSString *)number {
-    number = [BTUIKUtil stripNonDigits:number];
-    if (number.length == 0) {
-        return [[self class] allCards];
-    }
-    NSMutableSet *possibleCardTypes = [NSMutableSet set];
-    BTUIKCardType *card = [self cardTypeForNumber: number];
-    if (card) {
-        [possibleCardTypes addObject:card];
-    }
-    return [possibleCardTypes allObjects];
-}
-
 #pragma mark - Instance methods
 
 - (BOOL)validCvv:(NSString *)cvv {
--- a/BraintreeUIKit/Models/BTUIKCardType.m
+++ b/BraintreeUIKit/Models/BTUIKCardType.m
@@ -63,13 +63,13 @@
         return nil;
     }
     for (BTUIKCardType *cardType in [[self class] allCards]) {
-        for (NSString *prefix in cardType.validNumberPrefixes) {
-            if (number.length >= prefix.length) {
-                NSUInteger compareLength = MIN(prefix.length, number.length);
-                NSString *sizedNumber = [number substringToIndex:compareLength];
-                if ([sizedNumber isEqualToString:prefix]) {
-                    return cardType;
-                }
+        for (NSString *pattern in cardType.validNumberPrefixes) {
+            NSError  *error = nil;
+            NSRegularExpression* regex = [NSRegularExpression regularExpressionWithPattern:pattern options:0 error:&error];
+            NSTextCheckingResult *match = [regex firstMatchInString:number options:0 range: NSMakeRange(0, [number length])];
+
+            if (match != nil) {
+                return cardType;
             }
         }
     }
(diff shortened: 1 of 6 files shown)


BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -84,17 +84,9 @@ + (NSArray *)possibleCardTypesForNumber:(NSString *)number {
...
-            NSString *sizedPrefix = [prefix substringToIndex:compareLength];
-            NSString *sizedNumber = [number substringToIndex:compareLength];
-            if ([sizedNumber isEqualToString:sizedPrefix]) {
-                [possibleCardTypes addObject:cardType];
-                break;
-            }
-        }
+    BTUIKCardType *card = [self cardTypeForNumber: number];
+    if (card) {
+        [possibleCardTypes addObject:card];
```
bluk: Should this change repeat the new logic in lines 66-73 but don't do an early return? (So it should keep adding cards instead of just the first one). For the cases where this an overlap in the regular expressions.

demerino: Technically it should - yes. I think that we might want to just remove this method altogether. It is only used in one location in `BTUIKCardNumberFormField.m`:

```
- (BOOL)isPotentiallyValid {
    return [BTUIKCardType possibleCardTypesForNumber:self.number].count > 0;
}
```

So I just returned the first one since it would preserve the logic. Should we just rip it out and change the logic to:

```
- (BOOL)isPotentiallyValid {
    return [BTUIKCardType cardTypeForNumber:self.number] != nil;
}
```

I believe we need this method before to compare our hardcoded prefixes to see if if a short string (1-3 chars) COULD be valid. I think the new regex bypasses that.

bluk: I think some of the card bins could overlap each other. However, if the method isn't currently really used, I think the change you suggested could be done and if we need to go back and add more valid cards, we could bring it back if it makes maintenance easier.

demerino: Right, to check if something `isPotentiallyValid` - all we care about is *at least one*. I'll make that change.

demerino: Updated

BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -63,40 +63,17 @@ + (instancetype)cardTypeForNumber:(NSString *)number {
...
-                if ([sizedNumber isEqualToString:prefix]) {
-                    return cardType;
-                }
-            }
-        }
-    }
-    return nil;
-}
+        for (NSString *pattern in cardType.validNumberPrefixes) {
+            NSError  *error = nil;
```
skunkworks: `NSError *error;`, no need to set explicitly to `nil`.

BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -63,40 +63,17 @@ + (instancetype)cardTypeForNumber:(NSString *)number {
...
+            NSRegularExpression* regex = [NSRegularExpression regularExpressionWithPattern:pattern options:0 error:&error];
```
skunkworks: `NSRegularExpression *regex`

BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -63,40 +63,17 @@ + (instancetype)cardTypeForNumber:(NSString *)number {
...
-    for (BTUIKCardType *cardType in [[self class] allCards]) {
-        for (NSString *prefix in cardType.validNumberPrefixes) {
-            NSUInteger compareLength = MIN(prefix.length, number.length);
-
-            NSString *sizedPrefix = [prefix substringToIndex:compareLength];
-            NSString *sizedNumber = [number substringToIndex:compareLength];
-            if ([sizedNumber isEqualToString:sizedPrefix]) {
-                [possibleCardTypes addObject:cardType];
-                break;
+            if (match != nil) {
```
skunkworks: Might have to check `match.location != NSNotFound`

BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -134,37 +111,37 @@ + (NSArray *)allCards
 
         BTUIKCardType *visa = [[BTUIKCardType alloc] initWithBrand:BTUIKLocalizedString(CARD_TYPE_VISA)
                                                   securityCodeName:BTUIKLocalizedString(CVV_FIELD_PLACEHOLDER)
-                                                          prefixes:@[@"4"]];
+                                                          prefixes:@[@"^4\\d*"]];
```
skunkworks: I might consider removing or deprecating `validNumberPrefixes` in favor of `binRangePattern` which can be the regex pattern string.

demerino: I also debated this. I kinda like the idea of having multiple regex patters (as a possibility). If we ever needed it, it could aid with the readability of the regex patterns. But in either case, `prefixes` isn't quite right. But it isn't _totally_ incorrect since the regex primary search on bin ranges (prefixes).

> The leading six digits of the card number is the "issuer identification number (IIN)", sometimes referred to as the "bank identification number (BIN)".

BraintreeUIKit/Models/BTUIKCardType.m
demerino: I believe we talked about this in another PR. I'm going to move forward with this check. Also,`location` is not a property of the `match`.

Issue #32, opened 2017-03-24
Black square flickering over payment method selection popup


meismyles:
Issue when presenting the Drop-in controller - the payment selection view slides up and then has a black square that flickers over the payment options. This then seems to fade away behind the payment options, but is still visible behind them (its colour swapping between black and grey).

Seems to just randomly start appearing on devices at random times (happened on 2 different devices so far), at which point the only way to get rid of it is to restart the device. Force killing the app, or even deleting the app and reinstalling it does not solve it. Only restarting the device. This issue doesn't seem to occur when the app is first installed, but seems to just 'appear' randomly after a period of time.

Not sure if it could be something to do with memory? When presenting the Drop-in and then hitting cancel once it has loaded, the memory usage seems to increase by around 0.2/0.3 MB each time so not sure if there is a leak here somewhere?

The issue was occurring both plugged into the computer and not.

Screenshots attached for reference:

![1](https://cloud.githubusercontent.com/assets/6603912/24299085/e08b7878-109f-11e7-9033-dc4e8b38d77a.jpg)
![2](https://cloud.githubusercontent.com/assets/6603912/24299083/e07e5292-109f-11e7-868f-9b774eeb1aca.jpg)
![3](https://cloud.githubusercontent.com/assets/6603912/24299082/e07d3e66-109f-11e7-8ed2-119b834b937a.jpg)
![4](https://cloud.githubusercontent.com/assets/6603912/24299081/e07b3666-109f-11e7-9ee9-6d28d5695be3.jpg)

sdcoffey:
Hey @meismyles,

That is indeed strange. Neither I nor @demerino has seen this before. Would you mind running the demo app in this repository on that same device, and/or run your app on another device to determine whether this is an SDK or device issue?

meismyles:
It has occurred on 2 separate devices so far so it is not a device issue. When the issue was occurring on one of our devices (in our app), we opened the demo app in and it was **not** occurring in that.

However, I wouldn't say it eliminates the possibility that it exists within it. Maybe the constant use of our app and testing the integration within it brings it up? Very difficult to reproduce I know, it just seems to appear randomly, and then the only remedy is to restart the device.

meismyles:
It also seems to only occur when there are no recent/vaulted cards.

demerino:
I was able to create a similar square (it wasn't permanent though).

#39 is a potential fix - it looks like the animation of the loading spinner was causing some issues.

[...]

PR #39, opened 2017-04-04
Remove animation from loading spinner toggle.

demerino:
Potential fix for #32 

I've been able to reproduce a flickering square by reopening Drop-in multiple times (and using a client token with a customer). Removing the animation from the show/hide of the loading spinner rectifies the issue.


BraintreeDropIn/Public/BTDropInBaseViewController.h
```diff
@@ -34,8 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// Displays an overlay loading screen
 ///
 /// @param show Modifies the hidden property of the overlay
-/// @param animated Will animate the overlay changing the hidden property if set to `true`.
-- (void)showLoadingScreen:(BOOL)show animated:(BOOL)animated;
+- (void)showLoadingScreen:(BOOL)show;
```
bluk: Since this was public is it possible to deprecate the method and just add one without the animated property?

PR #56, opened 2017-10-04
added option to explicitly disable PayPal

hellmund:
This can be useful when having multiple merchants and currencies, with some supporting PayPal and some don't, and you want to disable PayPal globally as an option for certain users.

PR #66, opened 2017-11-22
Add Option to disable Venmo and PayPal

demerino:
This feature exists in the Andorid SDK. Inspired by https://github.com/braintree/braintree-ios-drop-in/pull/56

Also adding a private `BTDropInOverrides` class to assist in testing.


BraintreeDropIn/Models/BTDropInRequest.m
```diff
@@ -13,7 +13,10 @@ - (id)copyWithZone:(__unused NSZone *)zone {
     request.currencyCode = self.currencyCode;
     request.noShipping = self.noShipping;
     request.shippingAddress = self.shippingAddress;
+    request.additionalPayPalScopes = self.additionalPayPalScopes;
```
bluk: Good catch.

BraintreeDropIn/BTDropInOverrides.h
```diff
@@ -0,0 +1,8 @@
+#import <Foundation/Foundation.h>
+
+@interface BTDropInOverrides : NSObject
+
+/// Used for testing. Show Venmo option when configured correctly but Venmo is not installed.
+@property (class, nonatomic) bool displayVenmoOption;
```
bluk: Nitpick but should be `BOOL` instead of `bool` I think as well as in `BTDropInOverrides.m`.

PR #76, opened 2018-02-02
Update to 3DS Payment Flow

demerino:
Use `Braintree/PaymentFlow` integration for 3DS instead of `Braintree/3D-Secure`. **This will require a major version bump.**

- Enables 3DS Browser Switch
- Removes integration with `Braintree/3D-Secure`
- Use of `Braintree/3D-Secure` will no longer be recommended.

A nonce from Drop-in can be checked for 3DS status like so:
```
- (BOOL)nonceRequiresThreeDSecureVerification:(BTPaymentMethodNonce *)nonce {
    if ([nonce isKindOfClass:[BTCardNonce class]]) {
        BTCardNonce *cardNonce = (BTCardNonce *)nonce;
        return !cardNonce.threeDSecureInfo.wasVerified;
    }
    return false;
}
```

And additional 3DS verification can be performed outside of Drop-in:
```
- (void)performThreeDSecureVerification {
    BTAPIClient* apiClient = [[BTAPIClient alloc] initWithAuthorization: < YOUR AUTH STRING >];

    BTPaymentFlowDriver *paymentFlowDriver = [[BTPaymentFlowDriver alloc] initWithAPIClient:apiClient];
    paymentFlowDriver.viewControllerPresentingDelegate = self;

    BTThreeDSecureRequest *request = [[BTThreeDSecureRequest alloc] init];
    request.amount = [NSDecimalNumber decimalNumberWithString:@"10"];
    request.nonce = < YOUR NONCE >;
    [paymentFlowDriver startPaymentFlow:request completion:^(BTPaymentFlowResult * _Nonnull result, NSError * _Nonnull error) {
         self.selectedNonce = nil;
         if (error) {
             if (error.code == BTPaymentFlowDriverErrorTypeCanceled) {
                 // User cancelled
             }
             // Error
         }
        BTThreeDSecureResult *threeDSecureResult = (BTThreeDSecureResult *)result;
        // Now you can access the cardNonce
    }];
}
```


DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -7,7 +7,12 @@
 #import "BraintreeDemoSettings.h"
 #import "BTPaymentSelectionViewController.h"
 #import "BraintreeApplePay.h"
-#import "Braintree3DSecure.h"
+#import "BraintreeCard.h"
+#if __has_include("BraintreePaymentFlow.h")
```
skunkworks: Shouldn't be necessary to guard this, since it's only in the Demo app where it's guaranteed to be OK with `#import "BraintreePaymentFlow.h"`

Issue #72, opened 2017-12-15
Expiry Date Keyboard iPhone X Layout Issue


meismyles:
### General information

* SDK/Library version:  5.3.0
* Environment:  Sandbox & Production
* iOS Version and Device:  11.1.2 - iPhone X
* Integration type and version:  CocoaPods

### Issue description

I see you fixed the majority of iPhone X issues raised in #62 but the expiry date custom keyboard for the drop-in UI doesn't seem to be constrained to the safe area.

See screenshot below:
![braintreekeyboardissue](https://user-images.githubusercontent.com/6603912/34042203-d7a6924c-e19b-11e7-866e-4925f06c5c66.png)

demerino:
@meismyles Thanks for pointing this out, we'll work on this for an upcoming release.

demerino:
`6.1.0` should fix this issue.

Issue #62, opened 2017-11-16
iPhone X UI Support


shpuntov:
### General information

* SDK/Library version: 5.2.3
* Environment: Production
* iOS Version and Device: iOS 11.1 on an iPhone X
* Integration type and version: CocoaPods 1.3.1

### Issue description

As an iPhone X user, I'd want `BTDropInController` to be displayed in a safe area:

![simulator screen shot - iphone x - 2017-11-16 at 18 36 31](https://user-images.githubusercontent.com/1327944/32896669-1717b272-cafd-11e7-8e86-be8f4cd967f8.png)

demerino:
@shpuntov Thanks for pointing out this issue. We'll look into getting this fixed.

shpuntov:
Also just noted — **Cancel** navigation button in this dialog isn't properly aligned. I remove UI customization in my code and get the same: http://take.ms/RE08R
Sorry for adding this here — I'm not sure it's worth a separate issue. @demerino

demerino:
Got it. We'll look into this issue as well. Glad you're making use of the customization, looks good!

demerino:
PR: https://github.com/braintree/braintree-ios-drop-in/pull/65

demerino:
Should be fixed in 5.3.0

PR #79, opened 2018-02-19
UI Fixes

shpuntov:
This PR contains the following fixes and improvements:

- Fix for #72 
- "Better" fix for #62 (when `BTDropInController` loads payment methods, bottom inset still doesn't consider safe area on iPhone X)
- Fix `BTUIKBarButtonItem` font glitches on changing enabled state (when use custom fonts via `BTUIKAppearance` engine)
- Consider `BTUIKAppearance` UI customization for `BTUIKExpiryInputView` control

### DIFF
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -203,10 +203,10 @@
                                                                         views:viewBindings]];
     
     
-    self.contentHeightConstraint = [NSLayoutConstraint constraintWithItem:self.contentView attribute:NSLayoutAttributeTop relatedBy:NSLayoutRelationEqual toItem:self.view attribute:NSLayoutAttributeTop multiplier:1 constant:0];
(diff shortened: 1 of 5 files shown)


demerino:
@shpuntov This is great. Thank for taking the time to create this PR. We'll update once we've reviewed.

demerino:
@shpuntov I've walked through the various commits and the changes look appropriate. The new styling inheritance should align with the previous defaults.

I was curious what the `Prevent jump on loading BTDropInController` commit was referencing - can you describe what the jump was? I was having trouble recreating it.

shpuntov:
@demerino `Prevent jump on loading BTDropInController` commit fixes the following behavior: http://take.ms/nNVxE
Please note that bottom border of `BTDropInController` is displayed in unsafe area while payment methods are loaded — after load, `BTDropInController` arises over unsafe area.

PR #80, opened 2018-02-20
Enable secureTextEntry for BTUIKSecurityCodeFormField

demerino:
Update the `BTUIKSecurityCodeFormField` to `secureTextEntry = YES`.

cvv-masked


crookedneighbor:
I don't know my obj c well enough to tell, but is this opt in? Or will it automatically be applied for everybody?

demerino:
It will be auto applied. I didn't consider it to be breaking.

PR #81, opened 2018-02-20
Remove duplicates from BTUIKCardListLabel

demerino:
Ensure that duplicates are removed from the `BTUIKCardListLabel` when setting the `availablePaymentOptions`.

### DIFF
--- a/BraintreeUIKit/Components/BTUIKCardListLabel.m
+++ b/BraintreeUIKit/Components/BTUIKCardListLabel.m
@@ -40,7 +40,7 @@
 }
 
 - (void)setAvailablePaymentOptions:(NSArray *)availablePaymentOptions {
-    _availablePaymentOptions = availablePaymentOptions;
+    _availablePaymentOptions = [[NSSet setWithArray:availablePaymentOptions] allObjects];
     if ([BTUIKViewUtil isLanguageLayoutDirectionRightToLeft]) {
         _availablePaymentOptions = [[_availablePaymentOptions reverseObjectEnumerator] allObjects];
     }
--- a/UnitTests/BTUIKCardListLabelTests.m
+++ b/UnitTests/BTUIKCardListLabelTests.m
@@ -1,0 +1,19 @@
+#import <XCTest/XCTest.h>
+
+#import "BTUIKCardListLabel.h"
+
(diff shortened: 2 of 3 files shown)


UnitTests/BTUIKCardListLabelTests.m
```diff
@@ -0,0 +1,19 @@
...
+@interface BTUIKCardListLabelTests : XCTestCase
+
+@end
+
+@implementation BTUIKCardListLabelTests
+
+- (void)test_cardListLabel_removesDuplicates {
+    BTUIKCardListLabel *cardListLabel = [[BTUIKCardListLabel alloc] init];
+    cardListLabel.availablePaymentOptions = @[@"Visa", @"Visa", @"Discover", @"Visa"];
+    XCTAssertTrue([cardListLabel.availablePaymentOptions count] == 2, @"The array should contain 1 instance of Visa and Discover." );
```
bluk: `XCTAssertEqual` instead of `XCTAssertTrue`? Maybe also `1 instance each of Visa and Discover`?

BraintreeUIKit/Components/BTUIKCardListLabel.m
```diff
@@ -40,7 +40,7 @@ - (UIImage *) imageWithView:(UIView *)view
 }
 
 - (void)setAvailablePaymentOptions:(NSArray *)availablePaymentOptions {
-    _availablePaymentOptions = availablePaymentOptions;
+    _availablePaymentOptions = [[NSSet setWithArray:availablePaymentOptions] allObjects];
```
bluk: Is there any concern we should unique the values when loading the configuration itself instead of only on this class for display?

skunkworks: That's a good question -- I'd be inclined to make the change at a higher level within Drop-in, but if we go with the assumption that we'll patch this on the server, I think this is an acceptable stopgap.

demerino:
>  but if we go with the assumption that we'll patch this on the server, I think this is an acceptable stopgap.

This was my thinking. I wanted to reduce the footprint/impact of the change as much as possible so I went as low as possible (it also lended itself well to testing).

Thanks for the 👀

Issue #84, opened 2018-03-12
CVV error is not displayed


Minishlink:
### General information

* SDK/Library version: 6.1.0
* Environment: Sandbox
* iOS Version and Device: iOS 10.3.1 on an iPhone 6
* Integration type and version: CocoaPods 1.2.1

### Issue description

Use the following card:
* Number: 4000 0000 0000 0002
* CVV: 200

This raises an error `Record not found`, and nothing is displayed on the Braintree UI.

Note that on Android, an error is successfully displayed in the drop-in UI.

crookedneighbor:
Thanks for the report, we'll look into this

demerino:
@Minishlink You are right that the behavior is different from Android and Web where the UI is updated rather than returning the error. We'll work on a fix for this to put it inline with the other SDKs.

demerino:
This should be fixed in 6.2.0: https://github.com/braintree/braintree-ios-drop-in/releases/tag/6.2.0

PR #85, opened 2018-03-12
Display alert when tokenization fails on card form.

demerino:
Fix for https://github.com/braintree/braintree-ios-drop-in/issues/84

Current behavior: When CVV/AVS verification is enabled and an error is returned on tokenization - the error is returned via the Drop-in callback and the user is not able to retry. This is divergent from the web and Android experiences where the user is alerted of the issue and prompted to retry.

New behavior: An alert is shown when tokenization fails, prompting the user to `Please review your information and try again.` After dismissing the alert they are returned to the card form.


demerino:
Screenshot of the alert.
screen shot of alert

Issue #57, opened 2017-10-07
If use with vault, how to delete payment method in drop-in UI?


coolbeet:
Currently I dont see a way to let user delete payment method by themselves, is this feature on the product road map?

Thanks!

demerino:
This feature is not on our roadmap at this time. You'll need to delete the payment method on the server side: https://developers.braintreepayments.com/reference/request/payment-method/delete/ruby

coolbeet:
So drop-in UI is just for adding payment method right? 

Technically, one app probably should only have one payment method page where you can both add and delete payment method, this puts drop-in UI in an awkward position that, you have to build your own fully functional payment management page, and after that, there is no need to use drop-in UI since its functionality are all included in the payment management page.

A lot of our users already complain about they could find any place can delete payment method, we have to delete it in the gateway or by api. 

I strongly recommend build payment method deletion functionality in the drop-in UI, probably just let user long press the payment method and delete, just like how you delete app in springboard.

Thanks!

demerino:
I agree that it is awkward and we'll discuss some possible solutions to allow for editing.

One idea we might investigate is for an editing screen that will return a flag along with the nonce that says the users WANTS to delete the payment option. But it would still be up to the developer to implement the removal.

We appreciate the feedback.

coolbeet:
@demerino Thanks!

carterbryden:
Any updates on this?

demerino:
I recently took a pass at this. Let me know what you think of this approach: https://github.com/braintree/braintree-ios-drop-in/pull/89

coolbeet:
@demerino Great job, I like it!

demerino:
**Update:** We're actively working on this feature.

Unholny:
That's great news. Any estimation on when it's going live?

demerino:
I'm hesitant to give an estimate - but we're aiming for sometime in mid June 🤞

demerino:
6.3.0 adds this functionality, set the new flag on the BTDropInRequest to enable

```
request.vaultManager = true
```

coolbeet:
@demerino Awesome!

demerino:
I'll leave this issue open for questions. Let us know what you think. Web and Android will be coming in the next weeks.

Nonepse:
Thanks for this update @demerino! I updated the Pod in my project, but when trying to delete a card, I get an infinite spinner, and the card is not actually deleted:

[...]

PR #89, opened 2018-03-17
Edit Vaulted Payment Methods

demerino:
Fix for https://github.com/braintree/braintree-ios-drop-in/issues/57

Adds the ability to display an `Edit Mode` UI.
- Optional edit mode (requires client tokens with customers)

How to enable and use:
```
dropInRequest.vaultManager = @YES;
```

Example:
screen shot 2018-07-09 at 11 58 57 am


tlynchlaw:
Do I need to load any new cocoa pods to use this: dropInRequest.vaultManager = @YES;

This is a very useful feature and I need to add it.

sestevens:
Hi @tlynchlaw! I would recommend taking a look at our [Getting Started Guide](https://developers.braintreepayments.com/start/overview) if you haven't already. If you're already using Drop-In, you don't need to add any additional pods in order to vault payment methods.

You can also contact [Braintree Support](https://help.braintreepayments.com/) if you need help integrating with Braintree.

PR #92, opened 2018-04-04
Czech language support

TomasKypry:
###  Added localization file for czech language.


demerino:
Thanks for submitting this PR. We'll take a look and see if we can come up with a solution that will allow for any custom language to be added without needing a PR for each.

demerino:
Can you review https://github.com/braintree/braintree-ios-drop-in/pull/106

Let us know if this approach works for you to support custom translations.

demerino:
Closing in favor of #106

PR #102, opened 2018-05-29
Add Catalan translation support

aldevet:
Retaking https://github.com/braintree/braintree-ios-drop-in/pull/92 PR's conversation, has there been any progress on the ability to add languages?

I have added Catalan support as I believe it would be beneficial to allow PR in the meantime. Thanks.


demerino:
Thanks for submitting the PR. We've created an alternate PR to support custom translations https://github.com/braintree/braintree-ios-drop-in/pull/106

Can you try it out and see if it works for you?

PR #106, opened 2018-05-31
Add support for custom translations

demerino:
There are 2 steps to add support for a custom translation.

1. Add the translation file to your project using the name `BTUI.strings`. This file should be located in an appropriate locale folder. E.g. `cs.lproj`.
2. Activate the custom translation(s) via the SDK before initializing Drop-in (only needs to be done once):
```
[BTUIKLocalizedString customTranslations:@[@"cs"]];
 ```

**NOTE** This translation will only be used if it is the device's first _preferredLanguage_.
```
NSString *language = [[NSLocale preferredLanguages] firstObject];
```

### CODE REVIEW

**Comment by aldevet** [comment]
Thanks for this addition. I tried it and I can confirm it works great. The only problem is that `[[NSLocale preferredLanguages] firstObject]` returns `cs-US` (considering Czech as an example here) where `US` is the region you have your device setup with (that was changed starting from iOS 9 I believe).

Would it be possible to just check the language only? So for example, if you specify a custom translation of `["cz"]` it would still match a device setup with `cz-US`'s locale. If we make this assumption, that way we can even specify a specific locale just by declaring `["cz-US"]`.

**Comment by demerino** [comment]
Thanks for trying it out. I'll update the PR shortly to match/check on only the `language designator`.

In the future we might add `region` support.

**Comment by demerino** [comment]
@aldevet Give it another try. It will now only use the language designator portion of the preferred language.

Via apple docs: https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html#//apple_ref/doc/uid/10000171i-CH15

Locale ID syntax | Examples | Description
-- | -- | --
[language designator] | enfr | An unspecified region where the language is used.
[language designator]_[region designator] | en_GBzh_HK | The language used by and regional preference of the user.
[language designator]-[script designator] | az-Arabzh-Hans | An unspecified region where the script is used.
[language designator]-[script designator]_[region designator] | zh-Hans_HK | The script used by and regional preference of the user.

**Comment by aldevet** [comment]
Just tried it and works like a charm. Probably not the best place to ask this on but will Android support custom translations anytime soon? Thank you for your help!

**Comment by demerino** [comment]
Great! Thanks for taking the time to test it and give feedback.

I'll look into the Android support (I thought it already worked). We'll also need to update our documentation once we merge.

**Comment by aldevet** [comment]
Sorry to bother but is there any updates on the Android support? Thanks!

**Comment by demerino** [comment]
We just looked into it and it looks like custom translations should just work by adding the translations to your app and naming the files correctly `strings.xml`. See here for an example file: https://github.com/braintree/braintree-android-drop-in/blob/master/Drop-In/src/main/res/values-es/strings.xml

Let us know if you're still having issues.

**Comment by demerino** [comment]
Got it - working on a release.

File: BraintreeUIKit/Localization/BTUIKLocalizedString.m [hunk 1]
```diff
@@ -19,6 +31,10 @@ + (NSString *)localizationTable {
 
 #pragma mark Localization helpers
 
++ (void)customTranslations:(NSArray *)locales {
```
skunkworks: I think this should be renamed to `setCustomTranslations:`

BraintreeUIKit/Localization/BTUIKLocalizedString.m
```diff
@@ -2,9 +2,21 @@
...
 
-+ (NSBundle *)localizationBundle {
+static NSArray *customTranslations;
 
++ (NSBundle *)localizationBundle {
     static NSString * bundleName = @"Braintree-UIKit-Localization";
+    if ([[NSLocale preferredLanguages] count] > 0) {
+        NSString *language = [[NSLocale preferredLanguages] firstObject];
+        // Ignore region portion of local ID
+        language = [[language componentsSeparatedByString:@"_"] firstObject];
```
skunkworks: Slight preference for chaining this into one statement -- I initially interpreted this as a bug where language was getting overwritten. Something like this:

```
language = [[[[language componentsSeparatedByString:@"_"] firstObject] componentsSeparatedByString:@"-"] firstObject];
```

BraintreeUIKit/Localization/BTUIKLocalizedString.m
```diff
@@ -2,9 +2,21 @@
...
+        language = [[language componentsSeparatedByString:@"-"] firstObject];
+
+        if (customTranslations && [customTranslations containsObject:language]) {
```
skunkworks: Could be simplified to `if ([customTranslations containsObject:language])`

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -244,7 +244,8 @@ - (void)tappedToShowDropIn {
     } else {
         [BTUIKAppearance darkTheme];
     }
-    
+    [BTUIKLocalizedString customTranslations:@[@"cs"]];
```
skunkworks: Is this a leftover from testing?

demerino: This is intentional - for demo testing.

PR #113, opened 2018-07-16
Remove custom type for postcode field - Fixes #112

### LINKED ISSUE #112: Postcode field should support alphanumerical input
### General information

* SDK/Library version: 6.3.0
* Environment: Production
* iOS Version and Device: Tested with iOS 11.4, any model
* Integration type and version: CocoaPods 1.5.0

### Issue description

The postcode field only accepts 0-9 characters. According to [the list of postal code notation schemes](https://en.wikipedia.org/wiki/List_of_postal_codes) it is pretty common to have letters in a post code.
For example `SW11 3BZ` in London.

I think `BTUIKAppearance.m` [l.35](https://github.com/braintree/braintree-ios-drop-in/blob/master/BraintreeUIKit/Helpers/BTUIKAppearance.m#L35) and [l.55](https://github.com/braintree/braintree-ios-drop-in/blob/master/BraintreeUIKit/Helpers/BTUIKAppearance.m#L55) are the culprits.
The keyboardType is set in `BTUIKPostalCodeFormField.m` [l.16](https://github.com/braintree/braintree-ios-drop-in/blob/master/BraintreeUIKit/Components/BTUIKPostalCodeFormField.m#L16). It should be removed so it falls back to the default type `UIKeyboardTypeDefault` which I believe is the most appropriate for postcodes.

The Android SDK doesn't have this issue.

Nonepse:
Without UIKeyboardTypeNumberPad, the type will fall back to the default UIKeyboardTypeDefault which is more appropriate for postcodes.

| **Before**  | **After**  |
|---|---|
|   |  |

### DIFF
--- a/BraintreeUIKit/Components/BTUIKPostalCodeFormField.m
+++ b/BraintreeUIKit/Components/BTUIKPostalCodeFormField.m
@@ -13,7 +13,6 @@
         self.textField.accessibilityLabel = BTUIKLocalizedString(POSTAL_CODE_PLACEHOLDER);
         self.formLabel.text = BTUIKLocalizedString(POSTAL_CODE_PLACEHOLDER);
         self.textField.placeholder = @"12345";
-        self.textField.keyboardType = [BTUIKAppearance sharedInstance].postalCodeFormFieldKeyboardType;
 
         self.textField.autocorrectionType = UITextAutocorrectionTypeNo;
         self.textField.autocapitalizationType = UITextAutocapitalizationTypeNone;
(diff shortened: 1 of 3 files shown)


burnto:
Thanks for your contribution. I'm concerned it is not backward compatible because it changes default behavior.

demerino:
@Nonepse 
Why not just set `postalCodeFormFieldKeyboardType = UIKeyboardTypeDefault ` before initializing `BTDropInController`?

Nonepse:
@demerino Yep that's what I ended up doing after your reply in the ticket, to avoid keeping a fork of the repo on my side.

But I still feel it's wrong. Braintree service is international, and a postcode is not always numerical-only like in North America. So this solution looks like overriding an issue IMO.
Also, it took me some time to spot the issue since I was testing on the Simulator without the software keyboard. So it might not be immediately visible to the developer.

Nonepse:
Yep that's probably the best approach 👍

Nonepse:
Should've went into https://github.com/braintree/braintree-ios-drop-in/releases/tag/7.0.0 ?

scannillo:
The summary of this conversation resulted in PR (https://github.com/braintree/braintree-ios-drop-in/pull/207). I am closing this PR because the new one is open which covers this issue.

PR #121, opened 2018-09-20
luhn validate unionpay cards

jackellenberger:
UnionPay cards used to be issued with luhn invalid card numbers. We have always rejected luhn invalid cards on the backend, so this just brings the front end in line with that behavior.

@braintree/team-dx

### DIFF
--- a/BraintreeUIKit/Models/BTUIKCardType.m
+++ b/BraintreeUIKit/Models/BTUIKCardType.m
@@ -246,13 +246,11 @@
 #pragma mark - Validation
 
 - (BOOL)validAndNecessarilyCompleteNumber:(NSString *)number {
-    return (number.length == self.validNumberLengths.lastIndex &&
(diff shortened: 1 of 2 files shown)


BraintreeUIKit/Models/BTUIKCardType.m
```diff
@@ -246,13 +246,11 @@ - (NSAttributedString *)formatNumber:(NSString *)input {
 #pragma mark - Validation
 
 - (BOOL)validAndNecessarilyCompleteNumber:(NSString *)number {
-    return (number.length == self.validNumberLengths.lastIndex &&
-            ([BTUIKUtil luhnValid:number] || [self.brand isEqualToString:BTUIKLocalizedString(CARD_TYPE_UNION_PAY)]));
+    return (number.length == self.validNumberLengths.lastIndex && [BTUIKUtil luhnValid:number]);
```
crookedneighbor: Is this logic identical to the raw sdk? Should we just pull it from the raw sdk instead?

demerino: We don't run Luhn validation in the raw sdk level, just at this level for UI feedback.

PR #123, opened 2018-10-05
Accept Cardholder Name

billwerges:
Added an option for accepting a cardholder name when entering credit card details and did some minor project file cleanup.


BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -284,7 +289,11 @@ - (void)configurationLoaded:(__unused BTConfiguration *)configuration error:(NSE
 
 - (void)updateRequiredFields {
     NSArray <NSString *> *challenges = [self.configuration.json[@"challenges"] asStringArray];
-    self.requiredFields = [NSMutableArray arrayWithArray:@[self.cardNumberField, self.expirationDateField]];
+    self.requiredFields = [NSMutableArray arrayWithObject:self.cardNumberField];
+    if (self.dropInRequest.cardholderNameSetting >= BTFormFieldOptional) {
```
demerino: `!= BTFormFieldDisabled` instead?

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -826,6 +837,16 @@ - (void)formFieldDidChange:(BTUIKFormField *)formField {
     }
 }
 
+- (BOOL)formFieldShouldReturn:(BTUIKFormField *)formField {
+    if (formField == self.cardholderNameField) {
+        [self advanceFocusFromField:formField];
+        return NO;
+    }
+    else {
```
quinnjn: This `else` case isn't required, we could just `return YES;` if we didn't fall into the `if`

Issue #124, opened 2018-10-16
Unable to build project using `carthage update --no-build`


KoCMoHaBTa:
### General information

* SDK/Library version: 6.5.0
* Environment: ?
* iOS Version and Device: iOS 12 on iPhone XS Simulator
* Integration type and version: Carthage 0.31.1

### Issue description

I'm using Carthage for dependency manager by integrating the project references to my project directly from the Carthage/Checkout folder.
When i setup the dependency - it cannot build successfully and throws errors.

The Carthage command i use is `carthage update --no-build` - this produces the Carthage/Checkouts from where i integrate all dependencies using project references.
I've also tried to open and build the project directly from Xcode - i get the same error.

For more information - look at the images blow.

![cartfile](https://user-images.githubusercontent.com/1181346/47013737-e38c1f00-d150-11e8-82de-ee176804bc21.png)
![cartfile resolved](https://user-images.githubusercontent.com/1181346/47013738-e424b580-d150-11e8-9578-03ca612c5c84.png)
![embedded_binaries](https://user-images.githubusercontent.com/1181346/47013739-e424b580-d150-11e8-97a6-2a78d73204fe.png)
![error_file](https://user-images.githubusercontent.com/1181346/47013740-e424b580-d150-11e8-9990-9431f6d12110.png)
![error](https://user-images.githubusercontent.com/1181346/47013741-e424b580-d150-11e8-8167-f0ccb17af4b1.png)
![folder](https://user-images.githubusercontent.com/1181346/47013742-e4bd4c00-d150-11e8-94c7-a9cfc93c8848.png)
![link_binary_with_libraries](https://user-images.githubusercontent.com/1181346/47013743-e4bd4c00-d150-11e8-8469-5b1fb1d3f07d.png)
![target_dependencies](https://user-images.githubusercontent.com/1181346/47013745-e4bd4c00-d150-11e8-8856-33671db184c9.png)
![braintree_error](https://user-images.githubusercontent.com/1181346/47013837-306ff580-d151-11e8-8907-4f981fd82295.png)

billwerges:
[...]

PR #126, opened 2018-10-22
Linked against braintree dependency frameworks from Carthage/Checkouts folder

KoCMoHaBTa:
This PR is related to Issue #124


billwerges:
Hi @KoCMoHaBTa, unfortunately these changes only work when the directory layout is exactly as you have it using Carthage without the build option. For other users of `Braintree Drop-In`, they will simply see an invalid project and framework references within the project.

Our currently supported methods of integration are using CocoaPods, and using Carthage with Carthage performing the builds as well. For users who would like to manually integrate the project, they will need to set up the framework dependencies based on their build environment.

KoCMoHaBTa:
Hi @billwerges ,

I'm just curios, which is the use case where the users are seeing invalid project and framework references? Does it prevent users from building the project?
I would like to explore it and have it in mind in the future.

As far as I'm aware, at the moment, there are 2 cases:
- using Carthage - which is fine 
- using CocoaPods - which as far as I understand, it doesn't care about the Current project because it builds its own based on the given classes.

PR #128, opened 2018-11-05
Changed from using current calendar (default) to gregorian calendar f…

czgarrett:
…or date input

This fixes an issue with the drop in where if a user has the Buddhist calendar selected, it would present Buddhist years.  Credit cards are always issued with Gregorian years.


czgarrett:
Closing this PR because it was included in a separate fix.

Issue #127, opened 2018-11-05
Users of non-Gregorian calendars have incorrect expiration dates


calvinl:
### General information

* SDK/Library version: 
6.5.0

* Environment: 
Production & Sandbox

* iOS Version and Device: 
iOS 12 at least

* Integration type and version: 

### Issue description

If a non-Gregorian calendar is being used, such as the Buddhist calendar, we see expiration dates in the Drop UI that are 548 years ahead of the current date. 

[This code is responsible](https://github.com/braintree/braintree-ios-drop-in/blob/3abdc4b70255c4bad2f0b3d829707815d01f3336/BraintreeUIKit/Components/BTUIKExpiryInputView.m#L32-L44) for choosing the current calendar and then fetching the year, then displaying the years + 20 more years for selection in the Drop UI. 

It looks something like this:

![img_4005](https://user-images.githubusercontent.com/490687/48012721-a3391480-e0f0-11e8-815f-59bbf01681b5.png)

The use case is that some users in Thailand are experiencing this issue because they are using the Buddhist calendar. As far as whether this is a bug or not -- I'm assuming most if not all banks are issuing credit cards with expirations dates on the Gregorian calendar, so would it make more sense to have the above referenced code to always use Gregorian rather than `currentCalendar`?

hollabaq86:
Hi @calvinl thanks for bringing this to our attention! We [merged a PR](https://github.com/braintree/braintree-ios-drop-in/pull/129) that fixes this, which will be included in the next release. If you need the fix before then, here's the [commit](https://github.com/braintree/braintree-ios-drop-in/pull/129/commits/1d63df3513b7689307ae41061dfec15b87623e63).

calvinl:
@hollabaq86 Thank you! Looks great.

PR #129, opened 2018-11-05
set calendar type as Gregorian and syntax for pointers

hollabaq86:
Fixes Issue #127.

In addition to this fix, we made some style updates to the `BTUIKExpiryInputView.m` file.

### DIFF
--- a/BraintreeUIKit/Components/BTUIKExpiryInputView.m
+++ b/BraintreeUIKit/Components/BTUIKExpiryInputView.m
@@ -9,14 +9,14 @@
 
 @interface BTUIKExpiryInputView ()
 
-@property (nonatomic, strong) NSArray* months;
-@property (nonatomic, strong) NSArray* years;
-@property (nonatomic, strong) UICollectionView* monthCollectionView;
-@property (nonatomic, strong) UICollectionView* yearCollectionView;
-@property (nonatomic, strong) UIView* verticalLine;
+@property (nonatomic, strong) NSArray *months;
+@property (nonatomic, strong) NSArray *years;
+@property (nonatomic, strong) UICollectionView *monthCollectionView;
+@property (nonatomic, strong) UICollectionView *yearCollectionView;
+@property (nonatomic, strong) UIView *verticalLine;
 @property (nonatomic) NSInteger currentYear;
 @property (nonatomic) NSInteger currentMonth;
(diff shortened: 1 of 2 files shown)


BraintreeUIKit/Components/BTUIKExpiryInputView.m
```diff
@@ -30,12 +30,12 @@ - (instancetype)initWithFrame:(CGRect)frame {
         self.months = @[@"01", @"02", @"03", @"04", @"05", @"06", @"07", @"08", @"09", @"10", @"11", @"12"];
         
         NSDate *currentDate = [NSDate date];
-        NSCalendar* calendar = [NSCalendar currentCalendar];
-        NSDateComponents* components = [calendar components:NSCalendarUnitYear|NSCalendarUnitMonth fromDate:currentDate];
+        NSCalendar *calendar = [NSCalendar calendarWithIdentifier:NSCalendarIdentifierGregorian];
```
crookedneighbor: Don't know much about the code ramifications here, is this a breaking change?

Podfile.lock
```diff
@@ -45,15 +45,15 @@ PODS:
   - Braintree/Venmo (4.17.0):
     - Braintree/Core
     - Braintree/PayPalDataCollector
-  - BraintreeDropIn (6.4.1):
-    - BraintreeDropIn/DropIn (= 6.4.1)
-  - BraintreeDropIn/DropIn (6.4.1):
+  - BraintreeDropIn (6.5.0):
```
crookedneighbor: Why are these changing?

BraintreeUIKit/Components/BTUIKExpiryInputView.m
hollabaq86: No, this shouldn't be a breaking change (checked in with @billwerges on this)

Podfile.lock
hollabaq86: `6.5.0` is the latest version of iOS Drop-In

crookedneighbor: Yes, but why are these changing in this PR and not before? Was that a mistake before?

hollabaq86: Ah, sorry for not clarifying. It isn't necessarily a mistake, but the `Podfile.lock` file always lags the current release because it's self-referencing. This PR contains a new version because cocoapods now has the latest version.

PR #132, opened 2018-12-07
Project cleanup

billwerges:
Fixed all warnings and analysis issues, migrated project files to latest version of Xcode, and updated the demo to no longer be pinned to an old version of AFNetworking.


crookedneighbor:
@demerino I think you're the only one other than @billwerges with enough context on this repo. Mind taking a look at these changes?

PR #133, opened 2019-01-14
Fix #122 - Check for card tokenization errors also when 3DS is enabled

Nonepse:
When 3DS is enabled, the completion callback of:
```objective-c
- (void)tokenizeCard:(nonnull BTCardRequest *)request
             options:(nullable NSDictionary *)options
          completion:(nonnull void (^)(BTCardNonce *_Nullable __strong, NSError *_Nullable __strong))completion;
```
doesn't check for errors.

Because of this, the `BTPaymentFlowDriver` starts the payment flow even if an error was returned:
```objective-c
- (void)startPaymentFlow:(nonnull BTPaymentFlowRequest<BTPaymentFlowRequestDelegate> *)request
              completion:(nonnull void (^)(BTPaymentFlowResult *_Nullable __strong, NSError *_Nullable __strong))completionBlock;
```

Triggering the following request:
```
https://api.sandbox.braintreegateway.com:443/merchants/[merchant_id]/client_api/v1/payment_methods/(null)/three_d_secure/lookup
                                                                                                   ^^^^^^
```
Which is 404'ing.

Using an invalid card number that triggers a 422 error results in this undefined behavior:

By checking for error earlier in the flow, we catch all cases, and the SDK behaves correctly:

PR #137, opened 2019-02-04
UI Test Updates & Drop In Legacy Removal

scannillo:
- Removed all references to Braintree Legacy Drop In.

- Added UI test for public PR #133. 

- Made UI tests expiration date dynamic.


BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -618,12 +618,10 @@ - (void)basicTokenization {
             self.navigationItem.rightBarButtonItem = addCardButton;
 
             if (error != nil) {
-                dispatch_async(dispatch_get_main_queue(), ^{
```
jackellenberger: Is this change related to drop in legacy?

scannillo: We already switched to the main thread at line 615 and didn't need to do it again.

PR #145, opened 2019-03-18
add card vault opt out option

hollabaq86:
Work In Progress

## What's changing

- Adds `showSaveCardToggle`, a boolean that when `true` adds a toggle to let the customer choose to save a credit card to a merchant's Vault.
- Adds `defaultValueForVaulting`, a boolean that allows the merchant to set whether or not to have the save card input pre-selected.


BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -433,6 +446,7 @@ - (void)resetForm {
     _collapsed = YES;
     self.unionPayEnabledMerchant = NO;
     self.cardNumberField.hidden = NO;
+    // TODO: reset vault switch
```
scannillo: Was this taken care of?

BraintreeUIKit/Localization/BTUIKLocalizedString.m
```diff
@@ -251,6 +251,11 @@ + (NSString *)CARDHOLDER_NAME_LABEL {
     return NSLocalizedStringWithDefaultValue(@"CARDHOLDER_NAME_LABEL", [self localizationTable], [self localizationBundle], @"Cardholder Name", @"CARDHOLDER_NAME_LABEL");
 }
 
+//TODO: Add localizations
```
scannillo: Looks like localizations were added, can remove TODO

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## Unreleased
+
+* Add ability to opt out of client side vaulting
```
scannillo: Mention that this involves a customer facing UI element as well.

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -408,13 +416,18 @@ - (void)setCollapsed:(BOOL)collapsed {
...
+            self.shouldVaultCardSwitchField.hidden = ![self shouldDisplaySaveCardToggle] || collapsed;
             
             [self updateFormBorders];
             [self updateSubmitButton];
         }];
     });
 }
 
+- (BOOL)shouldDisplaySaveCardToggle {
+    return self.dropInRequest.allowVaultCardOverride && self.apiClient.tokenizationKey == nil;
```
scannillo: To confirm, vaulting cannot occur without a `client_token` and cannot occur with a `tokenization_key` right?

BraintreeUIKit/Public/BTUIKAppearance.h
```diff
@@ -47,6 +47,10 @@
 @property (nonatomic) UIKeyboardType postalCodeFormFieldKeyboardType;
 /// The highlighted version of the `tintColor`
 @property (nonatomic, readonly, getter = highlightedTintColor) UIColor *highlightedTintColor;
+/// Tint color for UISwitch when in the on position
+@property (nonatomic, strong) UIColor *switchOnTintColor;
```
demerino: Do we want to have defaults for our light and dark themes for these values? See https://github.com/braintree/braintree-ios-drop-in/blob/master/BraintreeUIKit/Helpers/BTUIKAppearance.m#L38

BraintreeDropIn/Public/BTDropInRequest.h
```diff
@@ -58,6 +58,14 @@ typedef NS_ENUM(NSInteger, BTFormFieldSetting) {
 /// Defaults to false.
 @property (nonatomic, assign) BOOL vaultManager;
 
+/// Optional: Whether or not to vault the card upon tokenization. When set to `false` with `allowVaultCardOverride` set to `false`, then cards will not be vaulted.
+/// Defaults to true
+@property (nonatomic, assign) BOOL vaultCard;
+
+/// Optional: When true, the card form will include an option to let the customer decide not to vault the card.
```
demerino: Do we want to be explicit that this setting only works when using a client token with a customer?

hollabaq86: Good call out! Clarified the description accordingly.

BraintreeUIKit/Public/BTUIKAppearance.h
scannillo: I think the default iOS colors look fine on light mode. Do we want to change anything on dark mode?

Light mode
Dark mode

BraintreeUIKit/Helpers/BTUIKAppearance.m
```diff
@@ -33,6 +33,8 @@ + (void)lightTheme {
     sharedTheme.activityIndicatorViewStyle = UIActivityIndicatorViewStyleGray;
     sharedTheme.useBlurs = YES;
     sharedTheme.postalCodeFormFieldKeyboardType = UIKeyboardTypeNumberPad;
+    sharedTheme.switchThumbTintColor = [UIColor whiteColor];
```
billwerges: The color constants are all class properties, so `UIColor.whiteColor`. Same applies to the `greenColor` below.

scannillo: The rest of the defaults are using the function call syntax [instead of the constants](https://github.com/braintree/braintree-ios-drop-in/blob/e8167e3fe078dbac45dbfb262888b705fc1fe050/BraintreeUIKit/Helpers/BTUIKAppearance.m#L19-L35). Shall we change them all for this PR?

billwerges: Sure, let's just modernize them all.

scannillo: Awesome. [Done](https://github.com/braintree/braintree-ios-drop-in/pull/145/commits/f110d9ef3b2ccdb8a5789e2bb2a45ab11ce194de) 👍

UITests/BraintreeDropIn_UITests.swift
```diff
@@ -884,3 +884,160 @@ class BraintreeDropIn_Error_UITests: XCTestCase {
...
+
+        let saveCardSwitch = elementsQuery.switches["Save card"]
+        self.waitForElementToBeHittable(saveCardSwitch)
+
+        guard let switchValue = saveCardSwitch.value else {
+            XCTFail()
+            return
+        }
+
+        XCTAssertEqual("0", switchValue as! String)
```
sestevens: It looks like changing this line to:
```swift
XCTAssertEqual("0", saveCardSwitch.value as? String)
```
would allow us to delete the `guard` statement right above. If `saveCardSwitch.value` is nil (or can't be cast to a String), the assertion will fail.

UITests/BraintreeDropIn_UITests.swift
```diff
@@ -884,3 +884,160 @@ class BraintreeDropIn_Error_UITests: XCTestCase {
...
+        self.waitForElementToBeHittable(saveCardSwitch)
+
+        XCTAssertTrue(saveCardSwitch.exists)
+
+        guard let switchValue = saveCardSwitch.value else {
+            XCTFail()
+            return
+        }
+
+        XCTAssertEqual("1", switchValue as! String)
```
sestevens: Here's another spot where we could simplify this to:

```swift
XCTAssertEqual("1", saveCardSwitch.value as? String)
```
and delete the `guard`. Also, there seems to be a rogue empty line just below this one.

PR #150, opened 2019-04-25
Added analytic event for vaulted card

scannillo:
- We previously collected analytic events for tokenization on new card creation only. This PR adds an analytic event for choosing an already vaulted card.

Events:
- Start event "vaulted-card.appear": when all vaulted cards are shown
- Success event "vaulted-card.select": when a vaulted card is chosen

Will squash before merge.

### DIFF
--- a/BraintreeDropIn/BTPaymentSelectionViewController.m
+++ b/BraintreeDropIn/BTPaymentSelectionViewController.m
@@ -53,6 +53,8 @@
 
 @implementation BTPaymentSelectionViewController
 
+static BOOL _vaultedCardAppearAnalyticSent = NO;
+
 - (id)init {
     self = [super init];
     if (self) {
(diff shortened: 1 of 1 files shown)


scannillo:
> Does iOS Drop-in not have unit tests around analytics?
Currently, there are none.

scannillo:
> Does iOS Drop-in not have unit tests around analytics?

Currently, there are none.

billwerges:
Do we have this event in the Android and Web drop-in projects already?

scannillo:
> Do we have this event in the Android and Web drop-in projects already?

Here is a [PR for Android](https://github.com/braintree/braintree-android-drop-in/pull/130). This task is in progress for Web.

hollabaq86:
@demerino That's phase 2 😄.

PR #153, opened 2019-06-26
Fix iOS 13 Font Issues

sestevens:
On iOS 13 fetching the system font by name no longer works. This PR adds read-only `font` and `boldFont` properties to BTUIKAppearance and defaults them to system font and system bold font if the font family and bold font family are not specified.

**iOS 13 - Before**
-------------------
Simulator Screen Shot - iPhone Xʀ - 2019-06-27 at 09 04 02

**iOS 13 - After**
-------------------
Simulator Screen Shot - iPhone Xʀ - 2019-06-27 at 09 05 31

### DIFF
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -124,7 +124,10 @@
 #pragma mark - Setup
 
 - (void)setUpViews {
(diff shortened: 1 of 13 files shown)


sestevens:
@demerino Good idea! I added before and after screenshots to the description.

PR #154, opened 2019-06-27
Fix spacing issue on expiry input view on iOS 13

sestevens:
On iOS 13, the month and year collection view height changes after the initial layout, and we were not updating the cell heights to match the new collection view height. This caused the month collection view to scroll, and the year collection view to display fewer years at once. We fixed this by invalidating the collection views' layout every time `layoutSubviews` is called.

iOS 13 - Before
--------------
ios-13-expiry-view-before

iOS 13 - After 
--------------
iOS-13-expiry-view-after

### DIFF
--- a/BraintreeUIKit/Components/BTUIKExpiryInputView.m
+++ b/BraintreeUIKit/Components/BTUIKExpiryInputView.m
@@ -16,7 +16,6 @@
 @property (nonatomic, strong) UIView *verticalLine;
 @property (nonatomic) NSInteger currentYear;
 @property (nonatomic) NSInteger currentMonth;
-@property (nonatomic, strong) UIPageControl *pageControl;
 
 @end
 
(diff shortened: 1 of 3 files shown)


BraintreeUIKit/Components/BTUIKExpiryInputView.m
```diff
@@ -115,133 +103,143 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
-    cell.label.textColor = [BTUIKAppearance sharedInstance].primaryTextColor;
     
-    if (self.selectedMonth && self.selectedMonth < self.currentMonth && [cell getInteger] == self.currentYear) {
-        cell.userInteractionEnabled = false;
+    if (isDisabled) {
+        cell.userInteractionEnabled = NO;
         cell.label.textColor = [BTUIKAppearance sharedInstance].disabledColor;
+    } else {
+        cell.userInteractionEnabled = YES;
+        cell.label.textColor = [BTUIKAppearance sharedInstance].primaryTextColor;
```
scannillo: This function is so much more readable ❤️. Nice job!

billwerges:
* Yes, we tested in landscape mode to ensure we could safely remove the notification check for orientation changes.
* We tested back to iOS 10 to ensure compatibility. Ideally we would have also tested iOS 9, but neither one of us could get that simulator to run on our work laptops.

PR #158, opened 2019-07-16
Add support for BTThreeDSecureRequest and 3DS 2 verification

demerino:
Adds `threeDSecureRequest` to BTDropInRequest and the ability to specify 3DS 2 and additional customer information.


BraintreeDropIn/BTDropInController.m
```diff
@@ -15,6 +15,11 @@
 #import <BraintreeCard/BraintreeCard.h>
 #import <BraintreeUnionPay/BraintreeUnionPay.h>
 #endif
+#if __has_include("BraintreePaymentFlow.h")
+#import "BraintreePaymentFlow.h"
+#else
+#import <BraintreePaymentFlow/BraintreePaymentFlow.h>
+#endif
```
sestevens: Why do we do this `#if __has_include` dance? I know it's all over the place, but I'm not totally clear on what it's doing.

BraintreeDropIn/Public/BTDropInRequest.h
```diff
@@ -41,6 +41,11 @@ typedef NS_ENUM(NSInteger, BTFormFieldSetting) {
 /// Defaults to false.
 @property (nonatomic, assign) BOOL threeDSecureVerification;
 
+/// Optional: Enable 3DS verification and specify options and additional information. If no amount is set, the `BTDropInRequest` `amount` will be used.
+///
+/// Note: For 3DS 2 options, set `billingAddress`, `amount`, `email`, `mobilePhone` for best results.
```
scannillo: Instead of "for 3DS 2 options", can we clarify this comment? Maybe something along the lines of "To encourage 3DS 2.0 flows".

UnitTests/BTDropInRequestTests.m
```diff
@@ -11,6 +12,30 @@ @interface BTDropInRequestTests : XCTestCase
...
+    billingAddress.region = @"CA";
+    billingAddress.countryCodeAlpha2 = @"US";
+    billingAddress.postalCode = @"54321";
+    threeDSecureRequest.billingAddress = billingAddress;
+
+    BTThreeDSecureAdditionalInformation *info = [BTThreeDSecureAdditionalInformation new];
+    info.shippingAddress = billingAddress;
+    info.shippingMethodIndicator = @"GEN";
+    threeDSecureRequest.additionalInformation = info;
+
```
scannillo: To me this will read better if the 3DS related fields (`BTThreeDSecureRequest`, `BTThreeDSecurePostalAddress`, and `BTThreeDSecureAdditionalInformation `)  fields go below the regular `BTDropInRequest` fields. This would also be consistent with the order we test them in.

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -22,6 +22,7 @@ @interface BraintreeDemoDropInViewController () <PKPaymentAuthorizationViewContr
 @property (nonatomic, strong) UIButton *dropInButton;
 @property (nonatomic, strong) UIButton *purchaseButton;
 @property (nonatomic, strong) UISegmentedControl *dropinThemeSwitch;
+@property (nonatomic, strong) UISegmentedControl *threeDSecureSwitch;
```
scannillo: To be consistent with the Braintree Core SDK demo app, I feel that the toggle b/w 3DS 1, 3DS 2, and No 3DS should live in the settings on the Demo App instead of a segmented control.

BraintreeDropIn/BTDropInController.m
demerino: It supports the framework or module import integration styles.

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
cooperbarth: Addressed in new changes (0e933fc).

UnitTests/BTDropInRequestTests.m
cooperbarth: We can't move the BTDropInRequest instantiation block up because the setting of the 3DS fields requires the other code in this function to have already run. We could move line 39 up alone, but that feels a little unnecessary. Is moving this whole block what you meant, or are we misunderstanding this?

BraintreeDropIn.podspec
```diff
@@ -27,11 +27,11 @@ Pod::Spec.new do |s|
...
-    s.dependency "Braintree/Card", "~> 4.23"
-    s.dependency "Braintree/Core", "~> 4.23"
-    s.dependency "Braintree/UnionPay", "~> 4.23"
-    s.dependency "Braintree/PaymentFlow", "~> 4.23"
-    s.dependency "Braintree/PayPal", "~> 4.23"
+    s.dependency "Braintree/Card", "~> 4.25"
+    s.dependency "Braintree/Core", "~> 4.25"
+    s.dependency "Braintree/UnionPay", "~> 4.25"
+    s.dependency "Braintree/PaymentFlow", "~> 4.25"
+    s.dependency "Braintree/PayPal", "~> 4.25"
```
billwerges: These dependencies should all be bumped to at least `4.26.1` since anything earlier will have App Store submission issues.

BraintreeDropIn/BTDropInController.m
```diff
@@ -269,12 +281,56 @@ - (void)cardTokenizationCompleted:(BTPaymentMethodNonce *)tokenizedCard error:(N
...
+
+    BTThreeDSecureRequest *request = self.dropInRequest.threeDSecureRequest;
+    if (!request) {
+        request = [[BTThreeDSecureRequest alloc] init];
+    }
+    request.nonce = tokenizedCard.nonce;
+
+    if (!request.amount && self.dropInRequest.amount) {
+        request.amount = [[NSDecimalNumber alloc] initWithString:self.dropInRequest.amount];
+    }
```
billwerges: We're handling a nil `BTThreeDSequreRequest` up on line 84. Here we're handling that case and the nil amount case. Can these checks be consolidated?

DropInDemo/Demo Base/Settings/Settings.bundle/Root.plist
```diff
@@ -110,6 +110,34 @@
...
+            <array>
+                <integer>0</integer>
+                <integer>1</integer>
+                <integer>2</integer>
+            </array>
+            <key>DefaultValue</key>
+            <integer>0</integer>
+            <key>Titles</key>
+            <array>
+                <string>Off</string>
```
billwerges: I know that `Off` was part of the segmented control UI, but it feels like it's in the wrong place here. The `Off` scenario should probably be handled by the `BraintreeDemoTransactionServiceThreeDSecureRequiredStatus` setting.

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -261,6 +262,30 @@ - (void)tappedToShowDropIn {
         dropInRequest.payPalRequest = [[BTPayPalRequest alloc] initWithAmount:@"4.77"];
     }
 
+    if ([BraintreeDemoSettings threeDSecureRequiredVersion] == BraintreeDemoTransactionServiceThreeDSecureRequestedVersionLegacy) {
+        dropInRequest.amount = @"10.00";
+        dropInRequest.threeDSecureVerification = YES;
```
billwerges: This conflicts with the check for `[BraintreeDemoSettings threeDSecureRequiredStatus]` above, on line 240.

BraintreeDropIn.podspec
billwerges: It looks like `pod update` was run, which will update the dependencies (and Podfile.lock) for the Demo app only. This file needs to be changed manually to specify that we require a minimum of `4.26.1`.

UITests/BraintreeDropIn_UITests.swift
```diff
@@ -809,6 +809,163 @@ class BraintreeDropIn_ThreeDSecure_UITests: XCTestCase {
...
+
+        self.waitForElementToBeHittable(app.buttons["Settings"])
+        app.navigationBars.buttons["Settings"].tap()
+        if (app.switches["Use Tokenization Key"].isSelected) {
+            app.switches["Use Tokenization Key"].tap()
+        }
+
+        app.tables.cells.staticTexts["Required (3DS Transaction Option)"].tap()
+        app.tables.cells.staticTexts["Required"].tap()
+        app.navigationBars.buttons["Settings"].tap()
```
billwerges: We don't normally us UI automation to change settings when running tests, use launch arguments that are then checked in `BraintreeDemoAppDelegate.m`. When these settings are changed using the UI, the new value is written to disk and persisted between launches, so if the order of the tests change this could have unintended consequences.

We should add support for a new launch argument named something like `ThreeDSecureVersion2`. The in `BraintreeDemoAppDelegate.m` we would do

```
if ([[[NSProcessInfo processInfo] arguments] containsObject:@"-ThreeDSecureVersion2"]) {
    [[NSUserDefaults standardUserDefaults] setInteger:BraintreeDemoTransactionServiceThreeDSecureRequestedVersion2 forKey:BraintreeDemoSettingsThreeDSecureVersionDefaultsKey];
} else {
    [[NSUserDefaults standardUserDefaults] setInteger:BraintreeDemoTransactionServiceThreeDSecureRequestedVersionLegacy forKey:BraintreeDemoSettingsThreeDSecureVersionDefaultsKey];
}
```

This ensure that the value will be reset to v1 on each UI test launch so the ordering of the tests doesn't matter.

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
```diff
@@ -25,6 +30,7 @@ typedef NS_ENUM(NSInteger, BraintreeDemoTransactionServiceThreeDSecureRequiredSt
 + (NSString *)authorizationOverride;
 + (BOOL)useTokenizationKey;
 + (BraintreeDemoTransactionServiceThreeDSecureRequiredStatus)threeDSecureRequiredStatus;
++ (BraintreeDemoTransactionServiceThreeDSecureRequestedVersion)threeDSecureRequiredVersion;
```
sestevens: Shouldn't this be `threeDSecureRequestedVersion`? _Requested_ rather than _required_.

UITests/BraintreeDropIn_UITests.swift
billwerges: The above snippet isn't fully correct. It should probably be

```
if ([[[NSProcessInfo processInfo] arguments] containsObject:@"-ThreeDSecureVersion2"]) {
    [[NSUserDefaults standardUserDefaults] setInteger:BraintreeDemoTransactionServiceThreeDSecureRequestedVersion2 forKey:BraintreeDemoSettingsThreeDSecureVersionDefaultsKey];
} else if ([[[NSProcessInfo processInfo] arguments] containsObject:@"-ThreeDSecureVersion1"]) {
    [[NSUserDefaults standardUserDefaults] setInteger:BraintreeDemoTransactionServiceThreeDSecureRequestedVersionLegacy forKey:BraintreeDemoSettingsThreeDSecureVersionDefaultsKey];
}
```

And we would want to change any 3DS1 UI tests to include the `-ThreeDSecureVersion1` launch argument.

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
```diff
@@ -17,6 +17,11 @@ typedef NS_ENUM(NSInteger, BraintreeDemoTransactionServiceThreeDSecureRequiredSt
     BraintreeDemoTransactionServiceThreeDSecureRequiredStatusNotRequired = 2,
 };
 
+typedef NS_ENUM(NSInteger, BraintreeDemoTransactionServiceThreeDSecureRequestedVersion) {
```
sestevens: Anyone know if there's a particular reason we say `TransactionService` in these enum names? It makes them soo long, and I'm not sure what additional information it provides.

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.m
```diff
@@ -49,6 +51,11 @@ + (BraintreeDemoTransactionServiceThreeDSecureRequiredStatus)threeDSecureRequire
     return [[NSUserDefaults standardUserDefaults] integerForKey:BraintreeDemoSettingsThreeDSecureRequiredDefaultsKey];
 }
 
++ (BraintreeDemoTransactionServiceThreeDSecureRequestedVersion)threeDSecureRequiredVersion {
```
scannillo: I think Susan point it out earlier, but just commenting here to be safe. The `threeDSecureRequiredVersion` seems like it should be `requested` to me

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -187,7 +187,8 @@ - (void)updatePaymentMethodConstraints {
     }
     [newConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:[dropInButton]-(20)-[purchaseButton]-(20)-[dropinThemeSwitch]" options:0 metrics:nil views:viewBindings]];
     [newConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"H:|-[dropinThemeSwitch]-|" options:0 metrics:nil views:viewBindings]];
-    
+
+
```
sestevens: Extra line
```suggestion

```

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.m
```diff
@@ -3,6 +3,8 @@
 NSString *BraintreeDemoSettingsEnvironmentDefaultsKey = @"BraintreeDemoSettingsEnvironmentDefaultsKey";
 NSString *BraintreeDemoSettingsCustomEnvironmentURLDefaultsKey = @"BraintreeDemoSettingsCustomEnvironmentURLDefaultsKey";
 NSString *BraintreeDemoSettingsThreeDSecureRequiredDefaultsKey = @"BraintreeDemoSettingsThreeDSecureRequiredDefaultsKey";
+NSString *BraintreeDemoSettingsThreeDSecureVersionDefaultsKey =
+    @"BraintreeDemoSettingsThreeDSecureVersionDefaultsKey";
```
sestevens: Looks like lines 6 and 7 should just be one line.

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
cooperbarth: You're definitely right that the naming is weird, but this is probably outside the scope of this PR. We might create a card to fix that up later.

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
```diff
@@ -18,9 +18,8 @@ typedef NS_ENUM(NSInteger, BraintreeDemoTransactionServiceThreeDSecureRequiredSt
 };
 
 typedef NS_ENUM(NSInteger, BraintreeDemoTransactionServiceThreeDSecureRequestedVersion) {
-    BraintreeDemoTransactionServiceThreeDSecureRequestedVersionOff = 0,
-    BraintreeDemoTransactionServiceThreeDSecureRequestedVersionLegacy = 1,
-    BraintreeDemoTransactionServiceThreeDSecureRequestedVersion2 = 2,
+    BraintreeDemoTransactionServiceThreeDSecureRequestedVersionLegacy = 0,
+    BraintreeDemoTransactionServiceThreeDSecureRequestedVersion2 = 1,
```
scannillo: Why did we remove the "off" feature of 3DS in the demo app? It seems like we would still want to be able to use the drop-in demo app without 3DS flows at all.

cooperbarth: The "off" flow would just be the same as just using a non-3DS card with the required status as "Default" or "Optional".

scannillo: So in general I think the "3DS Required Option" in the settings app currently is very confusing. The name doesn't really describe well what it does. To my understanding, that setting changes a merchant GW setting to either require (or not) 3DS transactions. If you guys have time, I would re-word these labels, but also understand if there isn't time now!

And yes, using a non-3DS card will trigger a "non 3DS" flow, but it will create a`ThreeDSecureRequest` (yes, it will fall out of this flow). Do we want an option of using the demo app that solely creates a `DropInRequest` and doesn't initialize any 3DS logic?

PR #159, opened 2019-07-17
Updates for iOS 13 Dark Mode

scannillo:
### What we changed:
- Added a new `colorScheme` property to `BTUIKAppearance` that allows merchants to choose from light, dark or dynamic drop-in color schemes. 
- Updated demo app to showcase these options.
- Updated demo app itself to be compatible with dark mode on iOS 13.
- Deprecated `lightTheme` and `darkTheme` class methods on `BTUIKAppearance`.
--------
**⚠️ Don't merge until Xcode 11 & iOS 13 are released ⚠️**


BraintreeUIKit/Public/BTUIKAppearance.h
```diff
@@ -2,11 +2,17 @@
...
+    BTUIKColorSchemeDynamic API_AVAILABLE(ios(13.0))
+};
+
 /// Shared instance used by Form elements
 + (instancetype)sharedInstance;
 
-+ (void)darkTheme;
-+ (void)lightTheme;
++ (void)darkTheme DEPRECATED_MSG_ATTRIBUTE("Use colorScheme property instead");
++ (void)lightTheme DEPRECATED_MSG_ATTRIBUTE("Use colorScheme property instead");
```
scannillo: We think it makes sense to deprecate these class methods in favor of the `colorScheme` property. But we're open to not deprecating if anyone feels differently.

UnitTests/BTUIKAppearanceTests.m
```diff
@@ -1,13 +1,220 @@
...
+#import "UIColor+BTUIK.h"
+#import "BTDropInController.h"
 
 @interface BTUIKAppearanceTests : XCTestCase
 
 @end
 
 @implementation BTUIKAppearanceTests
 
+- (void)test_lightTheme_classMethod {
```
scannillo: We had a hard time coming up with good names for these tests to distinguish testing the deprecated class methods from the new `colorScheme` property. Open to suggestions!

Currently we have: 
- `test_lightTheme_classMethod` for the class method
- `test_lightTheme_enum` for the property
- `test_dynamicTheme_enum_usingSystemLightMode` for the dynamic color scheme property
- etc...

sestevens:
Since we aren't planning to merge this until closer to the time that iOS 13 is released, we're closing this PR and opening a draft PR instead. See https://github.com/braintree/braintree-ios-drop-in/pull/160/files.

Issue #152, opened 2019-06-18
How to dismiss drop in ui when click outside.


lp2008:
Hello,
I want to dismiss the drop in ui when click outside of the drop in ui.
right now I can only dismiss when cancel button is clicked.

![Simulator Screen Shot - iPhone Xʀ - 2019-06-18 at 18 54 29](https://user-images.githubusercontent.com/6837443/59683987-f03ea680-91fa-11e9-92e2-c14cf459274b.png)

cooperbarth:
I was able to accomplish this by adding a UITapGestureRecognizer to the main view that calls `func dismiss(animated flag: Bool, completion: (() -> Void)? = nil)` on the BTDropInController. 

```
let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(dismissView))
tapRecognizer.delegate = self
dropIn?.view.addGestureRecognizer(tapRecognizer)

@objc func dismissView() {
  if (dropIn != nil) {
    dropIn?.dismiss(animated: true, completion: nil)
    dropIn = nil
   }
}
```

Does this solution work for you?

lp2008:
@cooperbarth thanks a lot. it's working.

PR #165, opened 2019-08-02
Close Drop-In on tap outside

cooperbarth:
Adds the ability for a user to close the Drop-In on a touch outside the view. In many iOS subviews, the standard is to give the user the ability to close the view by touching outside of it, instead of only being able to perform this action through touching the "cancel" button. This also mirrors the behavior of the Android Drop-In.

This PR references [this issue](https://github.com/braintree/braintree-ios-drop-in/issues/152), in which I provided a workaround for this situation.

### DIFF
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -32,7 +32,7 @@
 @interface BTDropInControllerDismissTransition : NSObject <UIViewControllerAnimatedTransitioning>
 @end
 
-@interface BTDropInController () <BTAppSwitchDelegate, BTDropInControllerDelegate, BTViewControllerPresentingDelegate, BTPaymentSelectionViewControllerDelegate, BTCardFormViewControllerDelegate, UIViewControllerTransitioningDelegate, BTThreeDSecureRequestDelegate>
+@interface BTDropInController () <BTAppSwitchDelegate, BTDropInControllerDelegate, BTViewControllerPresentingDelegate, BTPaymentSelectionViewControllerDelegate, BTCardFormViewControllerDelegate, UIViewControllerTransitioningDelegate, BTThreeDSecureRequestDelegate, UIGestureRecognizerDelegate>
 
 @property (nonatomic, strong) BTConfiguration *configuration;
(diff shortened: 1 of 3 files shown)


sestevens:
The question about whether or not this is a breaking change is a good one. I don't _think_ it would break any merchant apps, unless they've implemented their own tap gesture recognizer to dismiss drop-in **AND** the method it triggers has some special logic beyond just dismissing the view controller. I think this new gesture recognizer would swallow up the tap, preventing the merchant's (hypothetical) method from being called.

cooperbarth:
I'm definitely leaning towards this counting as a breaking change, also due to what @sestevens mentioned about swallowing other UIGestureRecognizers. However, it might be ok if we add in documentation regarding this if it were released in the next major update?

crookedneighbor:
If there's any possibility of a breaking change (and it sounds like there is), we'll need to save this for a new major version, or have it be opt in (with it happening automatically in the next major version)

BraintreeDropIn/BTDropInController.m
```diff
@@ -32,7 +32,7 @@ @interface BTDropInControllerPresentTransition : NSObject <UIViewControllerAnima
 @interface BTDropInControllerDismissTransition : NSObject <UIViewControllerAnimatedTransitioning>
 @end
 
-@interface BTDropInController () <BTAppSwitchDelegate, BTDropInControllerDelegate, BTViewControllerPresentingDelegate, BTPaymentSelectionViewControllerDelegate, BTCardFormViewControllerDelegate, UIViewControllerTransitioningDelegate, BTThreeDSecureRequestDelegate>
+@interface BTDropInController () <BTAppSwitchDelegate, BTDropInControllerDelegate, BTViewControllerPresentingDelegate, BTPaymentSelectionViewControllerDelegate, BTCardFormViewControllerDelegate, UIViewControllerTransitioningDelegate, BTThreeDSecureRequestDelegate, UIGestureRecognizerDelegate>
```
sestevens: I don't think we need to use any `UIGestureRecognizerDelegate` methods.

scannillo:
Yeh, and you can't click on any of the payment methods, so we'll need a refactor. Fair to close this PR and then we can make a ticket for this work?

PR #166, opened 2019-08-02
Accessibility Font Size

cooperbarth:
This PR supports dynamic font sizes in the BTPaymentSelectionViewController and BTCardFormViewController classes. This is an important accessibility feature as per Apple's guidelines; before these changes, text would remain a static size when the system font was changed.

### DIFF
--- a/BraintreeUIKit/Helpers/BTUIKAppearance.m
+++ b/BraintreeUIKit/Helpers/BTUIKAppearance.m
@@ -125,7 +125,9 @@ static BTUIKAppearance *sharedTheme;
 
 + (void)styleAdjustsToSystemFontSize:(UILabel *)label {
     label.font = [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
-    label.adjustsFontForContentSizeCategory = YES;
+    if (@available(iOS 10.0, *)) {
+        label.adjustsFontForContentSizeCategory = YES;
+    }
 }
 
 + (UILabel *)styledNavigationTitleLabel {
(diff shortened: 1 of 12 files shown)


BraintreeUIKit/Helpers/BTUIKAppearance.m
```diff
@@ -123,6 +123,11 @@ + (void)styleSystemLabelSecondary:(UILabel *)label {
     label.textColor = [BTUIKAppearance sharedInstance].secondaryTextColor;
 }
 
++ (void)styleAdjustsToSystemFontSize:(UILabel *)label {
+    label.font = [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
+    label.adjustsFontForContentSizeCategory = YES;
```
billwerges: The `adjustsFontForContentSizeCategory` property was introduced in iOS 10, but drop-in still supports iOS 9, so this will crash on that version of iOS.

cooperbarth: Ran on iOS 9 simulator and can confirm that this does crash. Making updates.

sestevens:
I think we should allow merchants to opt-in to this behavior. If the rest of a merchant's app doesn't support dynamic type, it might be weird for the drop-in to support it. We could add a boolean that defaults to false, and only enables dynamic type when a merchant explicitly sets it to true.

cooperbarth:
> I think we should allow merchants to opt-in to this behavior. If the rest of a merchant's app doesn't support dynamic type, it might be weird for the drop-in to support it. We could add a boolean that defaults to false, and only enables dynamic type when a merchant explicitly sets it to true.
> 
> Also, I'm not sure if there's some alternative way to make dynamic type work on iOS 9, but that might be worth looking into. If not, we could make the boolean available in iOS 10 only, similar to what Sammy and I did [here](https://github.com/braintree/braintree-ios-drop-in/blob/58ee93cb4e0eab657a4fde953ff76a5b467bafce/BraintreeUIKit/Public/BTUIKAppearance.h#L8) for dark mode.

I don't know if I agree that making it opt-in is the right call. Regardless of whether or not the merchant's app supports Apple's accessibility best practices, I still feel like we should?

PR #167, opened 2019-08-15
Updated README to reflect 3DS 2.0 flow

cooperbarth:
Replaces the general 3DS 1.0 flow with the 3DS 2.0 in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,6 @@ The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure i
 ```swift
 let request = BTDropInRequest()
 request.threeDSecureVerification = true
-request.amount = "1.00"
 
 let threeDSecureRequest = BTThreeDSecureRequest()
 threeDSecureRequest.threeDSecureRequestDelegate = self
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the `BTDropInRequest` and set an amount. Then, create a `BTThreeDSecureRequest object`, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the `BTDropInRequest` before use.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the `BTDropInRequest` and set an amount. Then, create a `BTThreeDSecureRequest` object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the `BTDropInRequest` before use.
 
 ```swift
 let request = BTDropInRequest()
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ address.locality = "Chicago"
 address.region = "IL"
 address.postalCode = "12345"
 address.countryCodeAlpha2 = "US"
-request.billingAddress = address
+threeDSecureRequest.billingAddress = address
 
 // Optional additional information.
 // For best results, provide as many of these elements as possible.
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the `BTDropInRequest` and set an amount. Then, create a `BTThreeDSecureRequest object`, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the `BTDropInRequest` before use.
 
 ```swift
 let request = BTDropInRequest()
--- a/README.md
+++ b/README.md
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest() object and set as many fields on it as possible; the more fields that are set, the less likely it is for a user to be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
 
 ```swift
-let request =  BTDropInRequest()
+let request = BTDropInRequest()
 request.threeDSecureVerification = true
 request.amount = "1.00"
+
+let threeDSecureRequest = BTThreeDSecureRequest()
+threeDSecureRequest.threeDSecureRequestDelegate = self
+
+threeDSecureRequest.amount = 1.0
+threeDSecureRequest.email = "test@example.com"
+threeDSecureRequest.requested = .version2
+
+let address = BTThreeDSecurePostalAddress()
+address.givenName = "Jill"
+address.surname = "Doe"
+address.phoneNumber = "5551234567"
+address.streetAddress = "555 Smith St"
+address.extendedAddress = "#2"
+address.locality = "Chicago"
+address.region = "IL"
+address.postalCode = "12345"
+address.countryCodeAlpha2 = "US"
+request.billingAddress = address
+
+// Optional additional information.
+// For best results, provide as many of these elements as possible.
+let additionalInformation = BTThreeDSecureAdditionalInformation()
+additionalInformation.shippingAddress = address
+request.additionalInformation = additionalInformation
+
+request.threeDSecureRequest = threeDSecureRequest
 ```
 
 ### Managing payment methods
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest() object and set as many fields on it as possible; the more fields that are set, the less likely it is for a user to be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
```
hollabaq86:
```suggestion
The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest, too, and set an amount. Then, create a BTThreeDSecureRequest() object, setting as many additionalInformation fields as possible; the more fields that are set, the less likely you will encounter a decline or the user will be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
```

hollabaq86: This language feels a little too casual, I made some suggestions but I'm not sure if I'm 100% right mentioning `additionalInfo` please double check for accuracy.

README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
...
 ```swift
-let request =  BTDropInRequest()
+let request = BTDropInRequest()
 request.threeDSecureVerification = true
 request.amount = "1.00"
+
+let threeDSecureRequest = BTThreeDSecureRequest()
+threeDSecureRequest.threeDSecureRequestDelegate = self
+
+threeDSecureRequest.amount = 1.0
```
hollabaq86: The amount doesn't need to be two 0's for decimal places?

README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
...
+threeDSecureRequest.email = "test@example.com"
+threeDSecureRequest.requested = .version2
```
hollabaq86: just doublechecking because new to iOS -  there should be a `.` in `.version2`

cooperbarth: Yep, that's Swift shorthand for the enum value `BTThreeDSecureVersion2`.

README.md
cooperbarth: It's just a float, so that shouldn't matter. We could change the README to 1.00 so it's clearer if you'd prefer, but it's functionally equivalent.

crookedneighbor: Shouldn't this be a string when it's sent to the GW?

cooperbarth: From what I could find in the docs and the demo app, it looks like BTDropInRequest.amount is a string and BTThreeDSecureRequest.amount is an NSDecimalNumber.

scannillo: [Confirming](https://github.braintreeps.com/team-sdk/braintree-ios/blob/214c01b8db5fcb965f59ff3f7be56f1536ecf2ef/Demo/Features/3D%20Secure/BraintreeDemoThreeDSecurePaymentFlowViewController.m#L101) it is a `NSDecimalNumber`. Should could write something like ...

``` 
request.amount = [NSDecimalNumber decimalNumberWithString:@"10.32"];
``` 
to be the most clear of casting a string to a number.

cooperbarth: We could write `request.amount = NSDecimalNumber(string: "10.00")`, but I feel like that might be even more confusing than just changing this to 10.00.

scannillo: I don't feel too strongly either way. But whichever we use, number or string, should show two decimal places.

crookedneighbor: Definitely agree with @scannillo

cooperbarth: Updated! Would love 👀so we can get this merged.

README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
...
+address.givenName = "Jill"
+address.surname = "Doe"
+address.phoneNumber = "5551234567"
+address.streetAddress = "555 Smith St"
+address.extendedAddress = "#2"
+address.locality = "Chicago"
+address.region = "IL"
+address.postalCode = "12345"
+address.countryCodeAlpha2 = "US"
+request.billingAddress = address
```
scannillo: I think we should be setting this `billingAddress` on the `threeDSecureRequest` not the `request`.

crookedneighbor: Yeah, that's how we do it on web.

README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest() object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
 
 ```swift
```
crookedneighbor: Should there be an obj c version of this as well?

README.md
sestevens: So, I'm always confused about `request.amount` vs. `threeDSecureRequest.amount`, and I have a feeling merchants will be too. My understanding is that we want merchants to use `threeDSecureRequest.amount`, not `request.amount`. So, maybe we should just remove `request.amount` from this code snippet?

I know it's not part of this PR, but shouldn't we also mark `request.amount` as deprecated?

README.md
```diff
@@ -124,12 +124,39 @@ request.applePayDisabled = !canMakePayments
...
-let request =  BTDropInRequest()
+let request = BTDropInRequest()
 request.threeDSecureVerification = true
 request.amount = "1.00"
```
sestevens: So, I'm always confused about `request.amount` vs. `threeDSecureRequest.amount`, and I have a feeling merchants will be too. My understanding is that we want merchants to use `threeDSecureRequest.amount`, not `request.amount`. So, maybe we should just remove `request.amount` from this code snippet?

I know it's not part of this PR, but shouldn't we also mark `request.amount` as deprecated?

README.md
cooperbarth: Right, my bad. In the docs, we call the `threeDSecureRequest` `request`, and something must've gotten caught in the copypasta.

README.md
crookedneighbor: ^ That's what I did in the web version of this, so yeah, I think we should do that too to be consistent.

README.md
cooperbarth: It's the same process conceptually for both languages; do you mean adding ObjC Code snippets? The whole README is in Swift as of now.

sestevens: My guess is that anyone who uses Obj C will be able to translate the Swift snippets themselves. So I'd say we're good with just having Swift in the README.

README.md
sestevens: I'll open a separate PR to deprecate `request.amount`.

README.md
```diff
@@ -124,12 +124,38 @@ request.applePayDisabled = !canMakePayments
...
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest() object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
 
 ```swift
-let request =  BTDropInRequest()
+let request = BTDropInRequest()
 request.threeDSecureVerification = true
-request.amount = "1.00"
```
billwerges: This change should be reverted, and the `BTDropInRequest.amount` should definitely not be deprecated. The majority of merchants likely aren't using 3DS, and the `amount` value here is what most of them will be setting.

cooperbarth: I agree that it shouldn't be marked deprecated, but this code snippet does live in the 3DS-specific section of the README, so it should probably use the most up-to-date 3DS conventions, no?

billwerges: Never mind, I didn't realize this value was only used for 3DS. It can probably be removed from the doc's then.

crookedneighbor: Is the amount used for Apple Pay as well as 3ds?

README.md
```diff
@@ -124,12 +124,38 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest() object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
```
sestevens: This is super minor, but I don't think we need the parens on BTThreeDSecureRequest here.

```suggestion
The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the BTDropInRequest and set an amount. Then, create a BTThreeDSecureRequest object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
```

crookedneighbor: Can we put them in backticks?

cooperbarth: Done!

README.md
```diff
@@ -124,12 +124,38 @@ request.applePayDisabled = !canMakePayments
 
 ### 3D-Secure + Drop-In
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, then just enable it in the BTDropInRequest and set an amount.
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the `BTDropInRequest` and set an amount. Then, create a `BTThreeDSecureRequest object`, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the `BTDropInRequest` before use.
```
sestevens: The backticks should be around `BTThreeDSecureRequest`, not `BTThreeDSecureRequest object`, no?

```suggestion
The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it in the `BTDropInRequest` and set an amount. Then, create a `BTThreeDSecureRequest` object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the `BTDropInRequest` before use.
```

README.md
cooperbarth: ^ @Epreuve looked into this yesterday and couldn't find a conclusive answer; I think since the line isn't directly related to using 3DS 2.0, it's probably more important to get this merged than worry now (and we can always update it later).

PR #168, opened 2019-08-19
Deprecate BTDropInRequest `amount` property

sestevens:
## Summary
Deprecating the `amount` property on `BTDropInRequest`. Instead, the amount for 3DS flows should be set on `BTThreeDSecureRequest`.

### DIFF
--- a/DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
+++ b/DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
@@ -1,6 +1,6 @@
 #import <Foundation/Foundation.h>
 #import <BraintreeDropIn/BTDropInRequest.h>
-#import <BraintreePaymentFlow.h>
+#import <Braintree/BraintreePaymentFlow.h>
 
 extern NSString *BraintreeDemoSettingsEnvironmentDefaultsKey;
 extern NSString *BraintreeDemoSettingsCustomEnvironmentURLDefaultsKey;
--- a/DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
+++ b/DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
@@ -1,5 +1,6 @@
 #import <Foundation/Foundation.h>
 #import <BraintreeDropIn/BTDropInRequest.h>
+#import <BraintreePaymentFlow.h>
 
 extern NSString *BraintreeDemoSettingsEnvironmentDefaultsKey;
 extern NSString *BraintreeDemoSettingsCustomEnvironmentURLDefaultsKey;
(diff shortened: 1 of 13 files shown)


scannillo:
Not sure if we want to make a removal note around this [line](https://github.com/braintree/braintree-ios-drop-in/blob/9ae04fbb76325a6f8d5e1b98907c6b70126bc251/BraintreeDropIn/BTDropInController.m#L84). This is the code that falls back to use the `dropInRequest.amount` when the `BTThreeDSecureRequest` wasn't created by the merchant, but they did request 3DS.

sestevens:
Just realized we'll need to do something about the fact that we use the old `amount` in the demo app ([here](https://github.com/braintree/braintree-ios-drop-in/blob/9ae04fbb76325a6f8d5e1b98907c6b70126bc251/DropInDemo/Features/Drop%20In/BraintreeDemoDropInViewController.m#L261)). I'll take a look at the best way to work around that since we currently treat deprecation warnings as errors.

DropInDemo/Features/Drop In/BraintreeDemoDropInViewController.m
```diff
@@ -279,6 +278,11 @@ - (void)tappedToShowDropIn {
             threeDSecureRequest.email = @"test@example.com";
             threeDSecureRequest.shippingMethod = @"01";
             dropInRequest.threeDSecureRequest = threeDSecureRequest;
+        } else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+            dropInRequest.amount = @"10.00";
```
billwerges: If we're deprecating this value, why don't we just not use it in the demo app. Can't we create a `BTThreeDSecureRequest` object even if we're doing 3DS 1, and use the default `versionRequested` value?

sestevens: Yeah, that would definitely be better. I'll try that!

sestevens: Done ✅

UnitTests/BTDropInControllerTests.swift
```diff
@@ -0,0 +1,55 @@
...
+        
+        let threeDSecureRequest = BTThreeDSecureRequest()
+        threeDSecureRequest.amount = NSDecimalNumber(string: "100.34")
+        dropInRequest.threeDSecureRequest = threeDSecureRequest
+        
+        let dropInController = BTDropInController(authorization: "development_tokenization_key",
+                                                  request: dropInRequest,
+                                                  handler: nil)
+        
+        XCTAssertEqual(dropInController?.dropInRequest.threeDSecureRequest?.amount, NSDecimalNumber(string: "100.34"))
```
scannillo: Seems like this is testing that the `threeDSecureRequest.amount` overrides the `dropInRequest.amount`. The `doesNotOverrideThreeDSecureRequest` name in the title is confusing me. Can you clarify what the purpose of this test is?

DropInDemo/Demo Base/Settings/BraintreeDemoSettings.h
```diff
@@ -1,5 +1,6 @@
 #import <Foundation/Foundation.h>
 #import <BraintreeDropIn/BTDropInRequest.h>
+#import <BraintreePaymentFlow.h>
```
billwerges: This should be either `#import <BraintreePaymentFlow/BraintreePaymentFlow.h>` or `#import "BraintreePaymentFlow.h"`.

PR #170, opened 2019-08-21
Update release script

sestevens:
### Summary of changes

 - Update release script to clarify that pressing any key will cause the script to continue after pre-release checklist is shown.

 ### Checklist

 - [ ] ~Added a changelog entry~

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -148,7 +148,7 @@ namespace :release do
     say "* [ ] Tests (rake spec) are passing, manual verifications complete."
     say "* [ ] Email is composed and ready to send to braintree-sdk-announce@googlegroups.com"
 
-    abort(1) unless ask "Ready to release? "
+    abort(1) unless ask "Ready to release? Press any key to continue. "
   end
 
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -148,7 +148,7 @@ namespace :release do
     say "* [ ] Tests (rake spec) are passing, manual verifications complete."
     say "* [ ] Email is composed and ready to send to braintree-sdk-announce@googlegroups.com"
 
-    abort(1) unless ask "Ready to release? "
+    abort(1) unless ask "Ready to release? Press any key to continue. "
```
scannillo: Is it a ton more work to have the `Y` and `N` keys programmed to continue or cancel here?

sestevens: Probably not a ton, but it would be some work. I'd say let's leave this as a future improvement 🙂

sestevens: @scannillo Are you good with this being merged?

scannillo: YES!

PR #171, opened 2019-08-21
Show activity indicator on 3DS flow with vaulted payment method

sestevens:
### Summary of changes

 - Show an activity indicator when a 3DS flow is triggered by the user selecting a vaulted payment method. This prevents the user from tapping on the vaulted payment method multiple times, which caused the 3DS flow to fail b/c lookup was performed multiple times for the same nonce.
- Clean up semicolons and `self` in BraintreeDropIn_UITests. **NOTE:** This is a separate commit, and it might be easier to review it separately since there are a lot of small changes.

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -323,7 +323,6 @@
 }
 
 - (void)onLookupComplete:(__unused BTThreeDSecureRequest *)request result:(__unused BTThreeDSecureLookup *)result next:(void (^)(void))next {
-    [self.paymentSelectionViewController showLoadingScreen:NO];
     next();
 }
 
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -108,17 +108,12 @@
     [self.apiClient sendAnalyticsEvent:@"ios.dropin2.appear"];
 }
 
-- (void)viewDidAppear:(BOOL)animated {
-    [super viewDidAppear:animated];
-}
-
 - (void)viewWillDisappear:(BOOL)animated {
     [super viewWillDisappear:animated];
     [self.apiClient sendAnalyticsEvent:@"ios.dropin2.disappear"];
 }
 
-- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
-{
+- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
     [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
     // before rotating
     [coordinator animateAlongsideTransition:^(__unused id<UIViewControllerTransitionCoordinatorContext> context) {
@@ -311,7 +306,7 @@
 
     [paymentFlowDriver startPaymentFlow:request completion:^(BTPaymentFlowResult * _Nonnull result, NSError * _Nonnull error) {
         BTThreeDSecureResult *threeDSecureResult = (BTThreeDSecureResult *)result;
-
+        [self.paymentSelectionViewController showLoadingScreen:NO];
         if (self.handler) {
             BTDropInResult *dropInResult = [[BTDropInResult alloc] init];
             if (threeDSecureResult.tokenizedCard != nil) {
@@ -328,6 +323,7 @@
 }
 
 - (void)onLookupComplete:(__unused BTThreeDSecureRequest *)request result:(__unused BTThreeDSecureLookup *)result next:(void (^)(void))next {
+    [self.paymentSelectionViewController showLoadingScreen:NO];
     next();
 }
 
@@ -499,6 +495,7 @@
             result.paymentOptionType = type;
             result.paymentMethod = nonce;
             if ([BTUIKViewUtil isPaymentOptionTypeACreditCard:result.paymentOptionType] && [self.configuration.json[@"threeDSecureEnabled"] isTrue] && self.dropInRequest.threeDSecureRequest) {
+                [self.paymentSelectionViewController showLoadingScreen:YES];
                 [self threeDSecureVerification:nonce];
             } else {
                 self.handler(self, result, error);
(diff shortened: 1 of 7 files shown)


BraintreeDropIn/BTDropInController.m
```diff
@@ -328,6 +323,7 @@ - (void)threeDSecureVerification:(BTPaymentMethodNonce *)tokenizedCard {
 }
 
 - (void)onLookupComplete:(__unused BTThreeDSecureRequest *)request result:(__unused BTThreeDSecureLookup *)result next:(void (^)(void))next {
+    [self.paymentSelectionViewController showLoadingScreen:NO];
```
billwerges: I believe this is executed before the completion block of `startPaymentFlow` is called. Do we need hide the loading screen in both places?

BraintreeDropIn/BTPaymentSelectionViewController.m
```diff
@@ -420,9 +423,10 @@ - (CGFloat)collectionView:(__unused UICollectionView *)savedPaymentMethodsCollec
 - (void)collectionView:(__unused UICollectionView *)savedPaymentMethodsCollectionView didSelectItemAtIndexPath:(__unused NSIndexPath *)indexPath {
     BTUIPaymentMethodCollectionViewCell *cell = (BTUIPaymentMethodCollectionViewCell*)[savedPaymentMethodsCollectionView cellForItemAtIndexPath:indexPath];
     if (self.delegate) {
-        [self.delegate selectionCompletedWithPaymentMethodType:[BTUIKViewUtil paymentOptionTypeForPaymentInfoType:cell.paymentMethodNonce.type] nonce:cell.paymentMethodNonce error:nil];
+        [self.delegate selectionCompletedWithPaymentMethodType:[BTUIKViewUtil paymentOptionTypeForPaymentInfoType:cell.paymentMethodNonce.type]
+                                                         nonce:cell.paymentMethodNonce error:nil];
```
scannillo: To follow this pattern, should error be its own line? I know you didn't do that because it is a small amount of characters, but it might read more consistently.

UITests/BraintreeDropIn_UITests.swift
```diff
@@ -955,6 +955,49 @@ class BraintreeDropIn_ThreeDSecure_2_UITests: XCTestCase {
...
+        app.launchArguments.append("-EditModeCustomer")
+        app.launch()
+        sleep(1)
+        
+        waitForElementToBeHittable(app.buttons["Change Payment Method"])
+        app.buttons["Change Payment Method"].tap()
+    }
+    
+    func testDropIn_threeDSecure_2_challengeFlow_withVaultedPaymentMethod() {
+        waitForElementToBeHittable(app.collectionViews.cells.firstMatch)
```
scannillo: How do we know that the first vaulted payment method (if any) will be of the card payment type?

BraintreeDropIn/BTDropInController.m
scannillo: Isn't this accounted for in line 309?

BraintreeDropIn/BTPaymentSelectionViewController.m
sestevens: I just missed it. It should definitely go on its own line.

UITests/BraintreeDropIn_UITests.swift
sestevens: It's a little bit fragile, but we're setting customer ID to `123-edit-test` in the set up method and that customer ID has the vaulted payment method we want. It's not ideal b/c deleting the payment method would break these tests, but we have other tests that are also relying on a vaulted payment method existing for that customer ID.

BraintreeDropIn/BTDropInController.m
sestevens: There are a few different flows to consider:

1. Lookup completes successfully, but then the user hits cancel on the 3DS challenge screen, and they return to the payment selection screen. In that case, we don't want to show the spinner, so we hide it in `onLookupComplete`.

2. Lookup does not complete successfully, and `onLookupComplete` is not called. In that case, the `startPaymentFlow` completion block is called, so we need to hide the spinner at that point.

UITests/BraintreeDropIn_UITests.swift
scannillo: Gotcha, thank you for clarifying! The name "-EditModeCustomer" indicates to me that the vault manager is enabled and that a customer can delete payment methods in their vault.

... but we're just using it for its vaulted cards.

sestevens: Yeah, that's true. Should we change it to `-CustomerWithVaultedCards` or something like that?

BraintreeDropIn/BTDropInController.m
billwerges: I think that we call the `startPaymentFlow` completion block even if they cancel the 3DS flow, so just hiding it there should be sufficient.

UITests/BraintreeDropIn_UITests.swift
scannillo: So it seems to be currently used in[ other tests](https://github.com/braintree/braintree-ios-drop-in/blob/284eebbe4a9f0b6f970440311a73732dc283a1be/UITests/BraintreeDropinEditMode_UITests.swift#L17-L21) to see if `self.waitForElementToBeHittable(app.buttons["Change Payment Method"])`, or if vault manager is properly rendering.

So we could make it all encompassing to include the fact that 1) it has vaulted payments and 2) it shows edit mode.

`-CustomerEditVaultWithPayments`?

BraintreeDropIn/BTDropInController.m
sestevens: Hmmm... OK. I'll give that a try.

sestevens: Just pushed up that change 👍

UITests/BraintreeDropIn_UITests.swift
sestevens: It looks like the ability to edit vaulted payment methods is controlled by this flag: `-DisableEditMode`, not the customer ID. So I don't think we need to call out the edit mode stuff in the name of the customer ID flag.

sestevens: @scannillo I changed it to `-CustomerIDWithVaultedPaymentMethods`. Kinda wordy, but hopefully more descriptive? Lemme know if that works for you or if you have a better idea!

PR #185, opened 2019-09-24
Update sdk dependency

billwerges:
Updated minimum Braintree SDK requirement to 4.29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## Unreleased
+
+* Require Braintree ~> 4.29
+
 ## 7.4.0 (2019-08-29)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## Unreleased
+
+* Require Braintree ~> 4.29
```
scannillo: Should we mention why? For Pay with Venmo to work on iOS 13?

sestevens: Personally, I don't think we need to call it out. We can always look at the changelog for `braintree_ios` version 4.29.0.

scannillo: Right! Makes sense to not duplicate it, like we've discussed before. Thanks for the reminder.

PR #191, opened 2019-10-16
Update blur style used in dynamic color scheme

sestevens:
### Summary of changes

 - Update blur style used in dynamic color scheme to one that matches the light color scheme better

After testing this a bit more on a device, `UIBlurEffectSystemUltraThinMaterial` looks too dark when the device is in light mode. I think `UIBlurEffectSystemMaterial` matches our original light and dark themes more closely.

 ### Checklist

 - [ ] ~Added a changelog entry~ (we already have one for the dark mode changes)

- @sestevens 

### Screenshots

Proposed: `UIBlurEffectStyleSystemMaterial` - Light Mode

Image from iOS (2)

Proposed: `UIBlurEffectStyleSystemMaterial` - Dark Mode

Image from iOS (3)

Previous: `UIBlurEffectStyleSystemUltraThinMaterial` - Light Mode (this looks too dark to me)

Image from iOS (1)

Previous: `UIBlurEffectStyleSystemUltraThinMaterial` - Dark Mode

Image from iOS


scannillo:
Hi @hollabaq86. We messed around with your feedback a bit to increase contrast. This will take more time than expected. We made a JIRA ticket to add this to our Accessibility backlog.

Are you OK with approving the PR as-is for the first round of this feature?

sestevens:
I updated this (again) to use Apple's system colors. This allows users get higher contrast in the Drop-In by turning on the high contrast setting on their phones. It still looks very similar to our original light and dark themes.

scannillo:
@billwerges 

> The change to system colors seems to have caused a regression. When the device is in dark mode and the dynamic theme is selected, the payment method selection view controller has the same background color as the parent view controller.

If you test on a device, the distinction between the view controllers is visible. Does that look OK to you?

Issue #177, opened 2019-09-05
Show activity indicator during payment method selection flow


francybiga:
### General information

* SDK/Library version: 7.3.0
* iOS Version and Device: iOS 12
* Integration type and version: Cocoapods 1.7.0

### Issue description

It would be great to show an activity indicator to the user during long operations in the drop-in flow. In particular there are same operations which take a considerable amount of time:
* begin of Paypal flow (it tipically took at least 2 seconds to to the system alert which asks the user to open the paypal webpage; during this time the user can tap on the Paypal payment selection celle multiple times, causing more than one modal window to be shown one above the other causing great confusion at the end of Paypal flow (another Paypal modal is still visible…)
* Dismissal of drop-in view controller: after the modal view controller is dismissed, BTPaymentSelectionViewController takes a little before being dismissed. During that time the user could select again a payment method and this causes the previously selected method to be discarded
* 3DS verification flow takes several seconds to be shown, leaving the user with a tappable selection method view

Note that this is very similar to what has been implemented in PR #171 but not limited to vaulted payment methods.

sestevens:
Hi @francybiga, this seems like a reasonable suggestion to me! We'll take a look at implementing this.

billwerges:
Fixed in [8.1.1](https://github.com/braintree/braintree-ios-drop-in/releases/tag/8.1.1)

PR #192, opened 2019-10-22
Callback Handler: Disable User Interaction

demerino:
### Summary of changes
Show loading indicator and disable user interaction when calling the final callback so that the user can not take additional actions.

Addresses one of the bullet points in https://github.com/braintree/braintree-ios-drop-in/issues/177

 ### Checklist

 - [x] Added a changelog entry

- demerino


billwerges:
Issue fixed in https://github.com/braintree/braintree-ios-drop-in/pull/223

PR #201, opened 2019-12-03
Fix Incorrect `colorScheme` Docstring

scannillo:
### Summary of changes
 - Fix the docstring for `BTUIKAppearance.colorScheme` property. 
- Reflect that the property _is_ available on any iOS, but instead the `dynamicScheme` is only available on iOS 13. [`BTUIKColorScheme` definition](https://github.com/braintree/braintree-ios-drop-in/blob/f1376c9d681750cdf1d91b8e5ed5f827993dc4e5/BraintreeUIKit/Public/BTUIKAppearance.h#L5-L9 ).

### Authors
@scannillo


scannillo:
retest this please

Issue #193, opened 2019-10-31
Deprecated UIWebView API usage


tobynguyen:
### General information

* SDK/Library version: 7.4.0
* Environment: Sandbox/Production
* iOS Version and Device: above iOS 12
* Integration type and version: CocoaPods 1.7.4

### Issue description

App Store Connect sends warning after uploading the app, that it contains a deprecated API usage of UIWebView. (ITMS-90809: Deprecated API Usage)

I don't know it's a false positive warning from the App Store side or what, but I keep getting this warning 2 weeks ago.

sestevens:
Hi @tobynguyen. The Braintree SDK relies on a dependency called CardinalMobile that uses `UIWebView`. We should have a release of the Braintree SDK out soon that contains an updated version of CardinalMobile without `UIWebView`.

We'll post here once the release is out, and then you'll need to run `pod update` to get the latest version of the base Braintree SDK. You can also follow along with [this issue](https://github.com/braintree/braintree_ios/issues/481) in the base SDK repo.

vlasov:
Hey @sestevens Braintree pod has been updated 3 days ago..
https://github.com/braintree/braintree_ios/issues/481#issuecomment-554534034

sestevens:
Updating to the latest version of the base Braintree SDK (version 4.30.2) should resolve this issue for users of Braintree Drop-In. If you're using Cocoapods, you'll need to run `pod update`.

vlasov:
@sestevens `pod update` will not have any effect, because Braintree Drop-In SDK is configured to  use particular version of Braintree SDK. In `BraintreeDropIn.podspec` file you can see:
```s.dependency "Braintree/Card", "~> 4.29"
    s.dependency "Braintree/Core", "~> 4.29"
    s.dependency "Braintree/UnionPay", "~> 4.29"
    s.dependency "Braintree/PaymentFlow", "~> 4.29"
    s.dependency "Braintree/PayPal", "~> 4.29"```
You have to update *podspec* and release a new version on BraintreeDropIn cocoapod in order to resolve this issue

sestevens:
Hi @vlasov. The `~> 4.29` syntax pins to version 4.29 and higher, up to (but not including) version 5.0. Since the fix was released in version 4.30.2, `pod update` should allow you to update to this version.

PR #208, opened 2020-01-23
Pin to Base SDK with Cardinal UIWebView Fix

scannillo:
### Summary of changes 

- The `UIWebView` [warning/alert issue ](https://github.com/braintree/braintree-ios-drop-in/issues/193)was caused by the Cardinal SDK. This issue was resolved in version 4.30.2 of `braintree_ios`. 
- This PR specifies Drop-In to require `~> 4.31` of Braintree SDK. 
- This [syntax](https://guides.cocoapods.org/using/the-podfile.html#specifying-pod-versions)  specifies v4.31 and the versions up to 5.0 not including 5.0 and higher. This allows us to require the fix for `UIWebView` and prevent breaking changes from the next major version.

 ### Checklist

 - [X] Added a changelog entry

### Authors
- @scannillo 
- @sestevens

### DIFF
--- a/BraintreeDropIn.podspec
+++ b/BraintreeDropIn.podspec
@@ -27,11 +27,11 @@ Pod::Spec.new do |s|
     s.source_files  = "BraintreeDropIn/**/*.{h,m}"
     s.public_header_files = "BraintreeDropIn/Public/*.h"
     s.frameworks = "UIKit"
-    s.dependency "Braintree/Card", "~> 4.30"
-    s.dependency "Braintree/Core", "~> 4.30"
-    s.dependency "Braintree/UnionPay", "~> 4.30"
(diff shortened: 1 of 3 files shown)


hollabaq86:
Builds are failing with the following message:

```[!] CocoaPods could not find compatible versions for pod "InAppSettingsKit":```

scannillo:
> Builds are failing with the following message:
> 
> `[!] CocoaPods could not find compatible versions for pod "InAppSettingsKit":`

Yeh. CI for ios-drop-in hasn't worked at all yet. To fix these builds would require another workstream. Our recent timeboxed attempts to fix it, haven't been enough. You can confirm by pulling locally/ running the tests!

PR #212, opened 2020-02-05
Support passing the vault option to the venmo tokenization service

demerino:
### Summary of changes

- Adds the vaultVenmo option and passes the `vault` key to the tokenization service.
- Requires an update to BraintreeVenmo to support the new key

 ### Checklist

 - [ ] Added a changelog entry

Co-authored-by: demerino <david.eros.merino@gmail.com>

### DIFF
--- a/BraintreeDropIn/Public/BTDropInRequest.h
+++ b/BraintreeDropIn/Public/BTDropInRequest.h
@@ -71,6 +71,10 @@ typedef NS_ENUM(NSInteger, BTFormFieldSetting) {
 /// Defaults to false
 @property (nonatomic, assign) BOOL allowVaultCardOverride;
 
+/// Optional: Whether or not to vault the venmo upon tokenization, must be set to false when using a client token without a customer.
+/// Defaults to true
+@property (nonatomic, assign) BOOL vaultVenmo;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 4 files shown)


BraintreeDropIn/Public/BTDropInRequest.h
```diff
@@ -71,6 +71,10 @@ typedef NS_ENUM(NSInteger, BTFormFieldSetting) {
 /// Defaults to false
 @property (nonatomic, assign) BOOL allowVaultCardOverride;
 
+/// Optional: Whether or not to vault the venmo upon tokenization, must be set to false when using a client token without a customer.
```
scannillo: If a merchant forgets to set this to `false`, and they use a client token without a customerID, won't this still crash since we're defaulting to `true`?

scannillo:
```suggestion
/// Optional: Whether or not to vault the Venmo payment method upon tokenization, must be set to `false` when using a client token without a `customerId`.
```

demerino: It will result in an integration error and not a crash:

> This causes an error when attempting to use Venmo as a payment option as the SDK will error with "Failed to store Venmo Account in vault."

At that point, the integrator should note the error and fix the integration to set the flag to `false`. This approach was taken to avoid a breaking change.

Alternatively we could make a breaking change and default to `false` and make this part of the major version bump. In that case, tokenization will always work for initial integrations but merchants will need to opt-in to vaulting.

demerino:
Did a manual test using this branch and the Braintree branch with the fix and the `vaultVenmo` flag worked as intended.

Issue #209, opened 2020-01-24
'topLayoutGuide' is deprecated: first deprecated in iOS 11.0


mavris:
### General information

* SDK/Library version: 4.31.0
* Environment: Both
* iOS Version and Device: iOS 13.2.2
* Integration type and version:  CocoaPods

### Issue description
When I set the iOS Deployment Target to iOS 11.0 in Pods for BrainTreeDropIn I am getting the following errors:
```
'topLayoutGuide' is deprecated: first deprecated in iOS 11.0 - Use view.safeAreaLayoutGuide.topAnchor instead of topLayoutGuide.bottomAnchor
```

scannillo:
Hi. Thanks for pointing this out. For now, please ignore the warning until we release a version with this fix. We can update here once that happens.

mavris:
This was not a warning! It was an error

scannillo:
Hi @mavris. Our SDK supports a minimum deployment target of iOS 9. Why do you need to change the deployment target of the SDK itself, instead of just your app's target? Leaving the SDK's deployment target as-is should work fine within your app targeting iOS 11.

scannillo:
This was fix in [v8.0.0](https://github.com/braintree/braintree-ios-drop-in/releases/tag/8.0.0)

alexkendall:
@mavris I have this same issue. can't upgrade the package because of a dependency. Were you able to workaround this?

sestevens:
@alexkendall What version of Braintree Drop-in are you using? This issue should be fixed in version 8.1.0.

andrew-stupchuk:
@scannillo still throws an error in 8.2.0 for react-native-braintree-dropin-ui. any solutions for this?

scannillo:
`react-native-braintree-dropin-ui` must be using an old version of `braintree-ios-drop-in`. Braintree/PayPal doesn't officially support that react native SDK, so you should raise your concern with them, or integrate with `braintree-ios-drop-in` directly.

mlazari:
@scannillo I am also seeing this error with `react-native-braintree-dropin-ui` and `BraintreeDropIn (8.2.0)`. I created a minimal repro React Native project: https://github.com/mlazari/BraintreeDropInErrorRepro

There are 4 errors at these lines:
- https://github.com/braintree/braintree-ios-drop-in/blob/8.2.0/BraintreeDropIn/BTDropInController.m#L423
[...]

Issue #197, opened 2019-11-15
Deprecated API usage


WojciechTrzasko:
### General information

* SDK/Library version:  7.4.0
* Environment: Sandbox/Production
* iOS Version and Device: iOS 11 and above
* Integration type and version: CocoaPods 1.7.4

### Issue description

Hi,

First of all, I really love your product!

Because of `UIApplication` API deprecations introduced with iOS 13, we investigated our project and its dependencies and find out that Braintree DropIn uses two deprecated properties:
* [statusBarOrientation](https://developer.apple.com/documentation/uikit/uiapplication/1623026-statusbarorientation) - used in `BTUIKExpiryInputView` and in `BTDropInController`.
* [statusBarFrame](https://developer.apple.com/documentation/uikit/uiapplication/1622937-statusbarframe) - used in `BTDropInController`.

Because we are getting close to the release of our project, we would like to be sure that there won't be any surprise at the end. In such a case I would like to ask, do you plan to resolve those in the near future?

Best regards!

scannillo:
Hi @WojciechTrzasko. Thank you for bringing this to our attention. We don't have an ETA on when we can make these updates, but we welcome public PRs if you need these changes urgently!

Team SDK - @hollabaq86

WojciechTrzasko:
Hi @scannillo,
As we have almost the same conversation in the issue for another repository, this time I will try to short the message ;).

We are preparing our app for MVP release in the middle of December and this is a very crucial date for us. Do you think that you will be able to resolve this before that date?

scannillo:
This is not on our current roadmap, but we've added it to our backlog. I do not think we will have this prioritized by then, since it is not blocking your or other's development. If it is however, please let us know.

scannillo:
This was fixed in [v8.1.0](https://github.com/braintree/braintree-ios-drop-in/releases/tag/8.1.0)

Crylion:
@scannillo I just found this old issue, because I was googling the exact same issue, I just encountered it in `BTUIKViewUtil.m` 🤔 

<img width="1274" alt="Screenshot 2023-05-09 at 15 42 46" src="https://github.com/braintree/braintree-ios-drop-in/assets/776753/a508e3a5-907c-44a0-a119-64b471976f37">

Crylion:
@scannillo Thanks for your response 🙏
[...]

PR #215, opened 2020-03-04
Fix deprecated APIs

sestevens:
### Summary of changes

 - Address a deprecation warnings mentioned in #197 and #209, specifically for `statusBarOrientation`, `statusBarFrame`, `topLayoutGuide` and `bottomLayoutGuide`. We need to continue using these properties to support older deployment targets, but we can use the newer alternatives when available.

 ### Checklist

 - [x] Added a changelog entry

- @sestevens 
- @billwerges


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+
+* Fix several UIKit deprecation warnings
```
scannillo: Why not specify the ones we changed?

sestevens: It might be a bit too much detail for the changelog, but here's what it would look like:

```
* Fix UIKit deprecation warnings for UIApplication's `statusBarOrientation` and `statusBarFrame` properties and UIViewController's `topLayoutGuide` and `bottomLayoutGuide` properties
```

What do you think? Too much info for the changelog?

scannillo: I'm for it. That's just me!

BraintreeUIKit/Helpers/BTUIKViewUtil.m
```diff
@@ -241,19 +241,48 @@ + (BTUIKVectorArtView *)vectorArtViewForVisualAssetType:(BTUIKVisualAssetType)ty
...
+
++ (CGFloat)statusBarHeight {
+    CGFloat statusBarHeight = 0;
+    if (@available(iOS 13, *)) {
+        statusBarHeight = CGRectGetHeight([self activeWindowScene].statusBarManager.statusBarFrame);
+    } else {
+        statusBarHeight = CGRectGetHeight(UIApplication.sharedApplication.statusBarFrame);
+    }
+
+    return statusBarHeight;
```
scannillo:
```suggestion
    if (@available(iOS 13, *)) {
        return CGRectGetHeight([self activeWindowScene].statusBarManager.statusBarFrame);
    } else {
        return CGRectGetHeight(UIApplication.sharedApplication.statusBarFrame);
    }
```

PR #224, opened 2020-05-08
Run UI tests on Travis CI

sestevens:
### Summary of changes

- Remove old test targets from DropInDemo scheme
- Update Podfile.lock
- Fix a failing UI test

 ### Checklist

 - [ ] ~Added a changelog entry~

- @sestevens


UITests/BraintreeDropIn_UITests.swift
```diff
@@ -112,8 +112,9 @@ class BraintreeDropIn_securityCodeValidation_CardForm_UITests: XCTestCase {
...
         app.launch()
         sleep(1)
-        waitForElementToBeHittable(app.buttons["Change Payment Method"])
-        app.buttons["Change Payment Method"].tap()
+        let button = app.buttons.containing(NSPredicate(format: "label LIKE '* Payment Method'")).element
+        waitForElementToBeHittable(button)
+        button.tap()
```
sestevens, line 117: This test shouldn't really care whether there's already a vaulted payment, so instead of looking for the "Change Payment Method" button, I updated it to tap either "Change Payment Method" or "Add Payment Method".

.travis.yml
```diff
@@ -6,5 +6,5 @@ cache:
   cocoapods: true
 
 script:
-  - set -o pipefail && xcodebuild -workspace BraintreeDropIn.xcworkspace -scheme UnitTests -destination platform\=iOS\ Simulator,OS\=13.3,name\=iPhone\ 11 test | bundle exec xcpretty -c
+  - set -o pipefail && xcodebuild -workspace BraintreeDropIn.xcworkspace -scheme DropInDemo -destination platform\=iOS\ Simulator,OS\=13.3,name\=iPhone\ 11 build test | bundle exec xcpretty -c
```
scannillo, line 9: The additional `build` command is needed now to build the demo app?

sestevens: It's not strictly necessary for running the tests. It's just checking that we can actually build the demo app.

PR #226, opened 2020-05-15
Update Constraints for SwiftUI Support

### LINKED ISSUE #202: SwiftUI support
### General information

* SDK/Library version: 7.5.0
* Environment: Production
* iOS Version and Device: 11.2.1
* Integration type and version: Cocoapods

### Issue description

I am trying to set up payment with Braintree, but Braintree does not yet support SwiftUI so I have to integrate it with UIKit. I created a wrapper using `UIViewControllerRepresentable` and I am presenting it as a modal using the `sheet` function; however, it does not work as expected, it seems it is opening two modals.

The screen when I open the modal:[![enter image description here][1]][1]

Here's my wrapper:
```
import SwiftUI
import BraintreeDropIn

struct BTDropInRepresentable: UIViewControllerRepresentable {
    var authorization: String
    var handler: BTDropInControllerHandler
    
    init(authorization: String, handler: @escaping BTDropInControllerHandler) {
        self.authorization = authorization
        self.handler = handler
    }
    
    func makeUIViewController(context: Context) -> BTDropInController {
        let bTDropInController = BTDropInController(authorization: authorization, request: BTDropInRequest(), handler: handler)!
        return bTDropInController
    }
    
    func updateUIViewController(_ uiViewController: BTDropInController, context: UIViewControllerRepresentableContext<BTDropInRepresentable>) {
    }
}
```

Here is where I am trying to open the modal:
```
Button(action: {
    self.checkout = true
}) {
    HStack {
        Spacer()
        Text("Checkout")
            .fontWeight(.bold)
            .font(.body)
        Spacer()
    }
    .padding(.vertical, 12)
    .foregroundColor(.white)
    .background(Color.blue)
}.sheet(isPresented: self.$checkout) {
    BTDropInRepresentable(authorization: self.token!, handler:  { (controller, result, error) in
                       if (error != nil) {
                           print("ERROR")
                       } else if (result?.isCancelled == true) {
                           print("CANCELLED")
                       } else if result != nil {
                        print("SUCCESS")
                           // Use the BTDropInResult properties to update your UI
                           // result.paymentOptionType
                           // result.paymentMethod
                           // result.paymentIcon
                           // result.paymentDescription
            }
        controller.dismiss(animated: true, completion: nil)
    })
}
```

Does anyone have experience with Braintree in SwiftUI or with a similar situation? Am I doing something wrong or forgetting something? 
I know writing my own views for Braintree checkout is an option, but I'd like to avoid that.

Thanks!

  [1]: https://i.stack.imgur.com/rDsGb.png

sestevens:
### Summary of changes

 - Update constraints for SwiftUI support. The `safeAreaLayoutGuide` does not take into account navigation and tool bars until `viewDidAppear`. See [Apple docs](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide).
- We also removed a check for `isBeingPresented`, which was not returning true in our SwiftUI demo. It looks like it's unnecessary (according to [these docs](https://docs.microsoft.com/pl-pl/dotnet/api/uikit.uiviewcontroller.isbeingpresented?view=xamarin-ios-sdk-12)).
- This resolves #202.

NOTE: To test this change on a SwiftUI demo app, checkout the [swiftui-demo branch](https://github.com/braintree/braintree-ios-drop-in/tree/swiftui-demo).

 ### Checklist

 - [x] Added a changelog entry

- @sestevens 
- @scannillo

### DIFF
--- a/BraintreeDropIn/BTDropInController.m
+++ b/BraintreeDropIn/BTDropInController.m
@@ -93,12 +93,16 @@
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
 
-    if (self.isBeingPresented) {
-        [self.paymentSelectionViewController loadConfiguration];
-        [self resetDropInState];
-        [self loadConfiguration];
-    }
+    [self.paymentSelectionViewController loadConfiguration];
+    [self resetDropInState];
+    [self loadConfiguration];
     [self.apiClient sendAnalyticsEvent:@"ios.dropin2.appear"];
+}
+
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    self.contentViewTopConstraint.constant = [self calculateContentViewTopConstraintConstant];
+    self.contentViewBottomConstraint.constant = -[self sheetInset];
 }
 
 - (void)viewWillDisappear:(BOOL)animated {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## unreleased
 
-* Show activity indicator on payment method selection screen at the beginning of PayPal, Venmo and 3DS flows
+* Show activity indicator on payment method selection screen at the beginning of PayPal, Venmo and 3DS flows (resolves #177)
+* Update constraints for SwiftUI support (resolves #202)
 
 ## 8.1.0 (2020-04-01)
 


BraintreeDropIn/BTDropInController.m
```diff
@@ -93,14 +93,18 @@ - (void)viewDidLoad {
...
+    [self.paymentSelectionViewController loadConfiguration];
+    [self resetDropInState];
+    [self loadConfiguration];
     [self.apiClient sendAnalyticsEvent:@"ios.dropin2.appear"];
 }
 
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    self.contentViewTopConstraint.constant = [self calculateContentViewTopConstraintConstant];
+    self.contentViewBottomConstraint.constant = -[self sheetInset];
```
sshropshire, line 105: Hey do we adjust the constants because we're pinning our top and bottom constraints to the superview instead of the safe area layout guide?

sestevens: It's basically just a difference in how the view controller gets presented in SwiftUI vs. UIKit.

**UIKit**
`-[BTDropInControllerPresentTransition animateTransition:]` is called when the Drop-in comes on screen. That method calls `-[BTDropInController flexViewAnimated:]`, which has these lines in it:

```
self.contentViewTopConstraint.constant = [self calculateContentViewTopConstraintConstant];
self.contentViewBottomConstraint.constant = -[self sheetInset];
```

That's what ensures that the payment selection view controller (the little modal) is the correct height when it initially appears.

**SwiftUI**
`-[BTDropInControllerPresentTransition animateTransition:]` is never called. (If you want to animate the transition, you need to do it from SwiftUI.)

So in SwiftUI, we need to set `self.contentViewTopConstraint.constant` and `self.contentViewBottomConstraint.constant` somewhere else. We're using the safe area, which doesn't get calculated correctly until `viewDidAppear`, so that's when we set the constants on those two constraints.

Let me know if that makes sense!

sshropshire: Yeah that makes sense. I forgot it was a pop-up modal for a second.

scannillo:
> Makes sense to do these changes. As far as testing, did we look at multiple devices (and even an iPad) and also in landscape mode?

Yep, we tested both iPad and landscape mode!

PR #234, opened 2020-09-23
(v9) Bump min deployment target to iOS 12

scannillo:
### Summary of changes

 - Bump min supported deployment target to iOS 12+
- Remove code that was kept for supporting iOS 9-11
- Bump SDK version in Podspec to 9.0.0
- Left a _question_ in the PR regarding flaky PwPP UI tests

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


UITests/BraintreeDropIn_UITests.swift
```diff
@@ -552,45 +552,56 @@ class BraintreeDropIn_PayPal_UITests: XCTestCase {
...
         waitForElementToBeHittable(webviewElementsQuery.links["Proceed with Sandbox Purchase"])
-        
         webviewElementsQuery.links["Proceed with Sandbox Purchase"].forceTapElement()
-        
+
+        // Assert on demo
         waitForElementToAppear(app.staticTexts["bt_buyer_us@paypal.com"])
-        
         XCTAssertTrue(app.staticTexts["bt_buyer_us@paypal.com"].exists)
     }
```
scannillo, line 580: **Note:** previously these PayPal UI test weren't running at all when running on an iOS 11+ simulator. I took a stab at making them work. I got them working, though they are a bit buggy - they only work on simulators running iOS 13 or 14, not below for some reason.

Do we want to keep these tests and write a comment, or nix them entirely? What about adding an integration test for a PayPal success case instead?

sestevens: I'd be OK with keeping them. We'll probably be testing on iOS 13 and 14 for the most part anyway.

An integration test is another good idea, although I'm not sure exactly what that would look like, since the Drop-in code is so UI-heavy.

scannillo: > An integration test is another good idea, although I'm not sure exactly what that would look like, since the Drop-in code is so UI-heavy.

True - I think it would be boil down to calling the `BTPayPalDriver` directly, which is essentially the[ PayPal IntegrationTests in `braintree-ios`](https://github.braintreeps.com/team-sdk/braintree-ios/blob/master/IntegrationTests/BraintreePayPal_IntegrationTests.m).

scannillo: Hey @sestevens - I left these tests in and added a comment to their flakiness. I opted to not use the `if #available(iOS 13.0, *)` annotation since I think having them fail is better than silently not execute. Let me know what you think 👍

sestevens: Makes sense to me!

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+
+* Breaking changes (v9)
+  * Bump minimum supported iOS version to 11.0+
```
sestevens, line 6:
```suggestion
  * Bump minimum supported deployment target to iOS 11.0
```

BraintreeUIKit/Components/BTUIKExpiryInputView.m
```diff
@@ -98,11 +98,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
                                                                      metrics:nil
                                                                        views:viewBindings]];
         
-        id bottomReferenceView = self;
-
-        if (@available(iOS 11.0, *)) {
-            bottomReferenceView = self.safeAreaLayoutGuide;
-        }
+        UILayoutGuide *bottomReferenceView = self.safeAreaLayoutGuide;
```
sestevens, line 101: We could just remove this variable and use `self.safeAreaLayoutGuide` directly. In my mind, that helps readability since I don't have to remember what the variable refers to.

If we want to keep the variable, how about renaming it to `bottomLayoutGuide` (since it's never actually a view anymore).

README.md
```diff
@@ -6,7 +6,7 @@ Welcome to Braintree's Drop-In SDK for iOS!
 
 ![Drop-in light theme](Docs/client-sdk-ios-series-light.png "Drop-in light theme")
 
-**The Braintree iOS Drop-In SDK requires Xcode 11+**. It permits a Deployment Target of iOS 9.0 or higher.
+**The Braintree iOS Drop-In SDK requires Xcode 12+**. It permits a Deployment Target of iOS 11.0 or higher.
```
sestevens, line 9:
```suggestion
**The Braintree iOS Drop-In SDK requires Xcode 12+**. It permits a deployment target of iOS 11.0 or higher.
```

sestevens:
As part of this PR, it might make sense to update our [Travis file ](https://github.com/braintree/braintree-ios-drop-in/blob/master/.travis.yml)to use Xcode 12 and an iOS 14 simulator.

sestevens (approved):

sshropshire (approved):

PR #236, opened 2020-10-04
Fix Braintree module imports

demalex:
Hi,

Thank you very much for great library!

I am integrating `BraintreeDropIn` using cocoapods and I started to face an issue when specific installation option is used.

Starting with [1.7.0](https://blog.cocoapods.org/CocoaPods-1.7.0-beta/) cocoapods allow to specify `generate_multiple_pod_projects` option in `Podfile`. When used, it helps to improve performance and reduce compile time.

Unfortunately `BraintreeDropIn` fails to compile with this option in `Podfile` because cocoapods generates separate project per dependency which results in separate modules. Eventually compilation fails because `Braintree` module imports are not correct. Please see attached logs: [Build BraintreeDropIn_2020-10-04T20-34-19.txt](https://github.com/braintree/braintree-ios-drop-in/files/5324435/Build.BraintreeDropIn_2020-10-04T20-34-19.txt)

I would be grateful for your feedback.

Thanks!

### Summary of changes

 - Fix `Braintree` module imports
 - Update changelog

 ### Checklist

 - [ ] Added a changelog entry

- @demalex

### DIFF
--- a/BraintreeDropIn/BTAPIClient_Internal_Category.h
+++ b/BraintreeDropIn/BTAPIClient_Internal_Category.h
@@ -1,7 +1,7 @@
 #if __has_include("BraintreeCore.h")
 #import "BraintreeCore.h"
 #else
-#import <BraintreeCore/BraintreeCore.h>
+#import <Braintree/BraintreeCore.h>
 #endif
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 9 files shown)


billwerges:
Hi @demalex,

Unfortunately these changes would break integrations for developers using Carthage and manual imports, so this is not something we will be able to accept at this time. We will investigate adding support for this CocoaPods feature in a future release.

RamblinWreck77:
Probably not worth it with SPM support on the horizon. I don't know anyone wanting to stay with cocoapods long term.

demalex:
Hi @billwerges,

I see. It might be worth to add a readme note regarding this use case.

Anyway, thanks for your feedback!

PR #246, opened 2020-12-02
Convert CI to GH Actions

scannillo:
### Summary
 - Remove travis file
 - Add .github/workflows/testing.yml file to run Unit and UI Tests

 ### Checklist

~Added a changelog entry~

### Authors
@scannillo

PR #247, opened 2020-12-02
Cleanup Gemfile

scannillo:
### Summary of changes

- Cleanup unused Ruby gems
- Bump Ruby version to match that in braintree_ios

### Question

- There is a chance some folks were using `git-pairing`, but I doubt it/haven't seen it. Comment if objections!

### Authors
@scannillo


sestevens (approved):

scannillo:
The travis build is failing because of the ruby version. I don't think this will be an issue when we're using [GH Actions in this PR.](https://github.com/braintree/braintree-ios-drop-in/pull/246) I am going to ignore that build fail for now.

sshropshire (approved):

PR #251, opened 2020-12-04
Updates for CI Stability

sestevens:
### Summary of changes

* Update GH Actions file so that we always use Xcode 12.2. UI tests were sometimes running into an error where GH Actions couldn't find the correct simulator, and this change to our `testing.yml` file seems to resolve that issue. See this [SO post](https://stackoverflow.com/a/59508420).

* Increase timeout on PayPal tests. These tests fail sporadically in CI, so maybe increasing the timeout will help?

* Bump Braintree pod to 4.36.1

 ### Checklist

 - ~Added a changelog entry~

- @sestevens

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -8,12 +8,12 @@ jobs:
       # Checks out a copy of your repository
       - name: Checkout repository
         uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


scannillo (approved):
🎸

PR #252, opened 2020-12-04
UI Test Improvements

sestevens:
### Summary of changes

* Tap in corner of screen during PayPal flow so that the tests don't
accidentally dismiss the PayPal site early

* Force tap into text fields to (hopefully) decrease flakiness

 ### Checklist

 - ~Added a changelog entry~

- @sestevens

### DIFF
--- a/UITests/BraintreeDropIn_UITests.swift
+++ b/UITests/BraintreeDropIn_UITests.swift
@@ -284,6 +284,7 @@ class BraintreeDropIn_CardholderNameAvailable_UITests: XCTestCase {
         cardNumberTextField.typeText("4111111111111111")
         
         let cardholderNameTextField = elementsQuery.textFields["Cardholder Name"]
+        cardholderNameTextField.forceTapElement()
         cardholderNameTextField.typeText("\n")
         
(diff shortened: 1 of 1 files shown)


UITests/BraintreeDropIn_UITests.swift
```diff
@@ -567,13 +570,13 @@ class BraintreeDropIn_PayPal_UITests: XCTestCase {
             }
             return true
         }
-        app.tap()
+        app.coordinate(withNormalizedOffset: CGVector.zero).tap()
```
sestevens, line 573: My theory is that `app.tap()` was usually tapping in the middle of the screen, right on the "Proceed with Sandbox Purchase" button on the PayPal site, which cause the flow to finish before the next lines of the test had a chance to run. If the tap occurs in the corner of the screen, we should be able to avoid this problem.

sshropshire (approved):

hollabaq86 (approved):
🚢

UITests/BraintreeDropIn_UITests.swift
scannillo: Awesome find!!!

PR #256, opened 2021-01-08
Spm support

scannillo:
### Summary of changes

 - Include modules `Braintree/ApplePay`, `Braintree/Venmo`, and `Braintree/UnionPay` in `BraintreeDropIn` for CocoaPods integrations.
 - Update project to use Carthage to fetch Braintree dependencies
     - Demo app, Unit Tests, & the BraintreeDropIn framework all fetch Braintree via Carthage
     - Update CI to also fetch deps via Carthage
     - **NOTE:** We tried to use SPM for our internal development, but it doesn't have enough features supporting local development
- Remove GET and POST methods from internal BTAPIClient header since [this PR in Braintree iOS](https://github.com/braintree/braintree_ios/pull/579) makes them public
    - This fixes an "ambiguous reference" error

 ### Checklist

 - [x] Added a changelog entry

### Authors
@scannillo @sestevens


Sources/BraintreeDropIn/BTDropInBaseViewController.m
```diff
@@ -1,8 +1,8 @@
-#import "BTDropInBaseViewController.h"
+#import <BraintreeDropIn/BTDropInBaseViewController.h> //syntax should be same for all since pod name is same as module name
```
sestevens, line 1:
```suggestion
#import <BraintreeDropIn/BTDropInBaseViewController.h>
```

Demo/Demo Base/DemoContainerViewController.swift
```diff
@@ -1,5 +1,5 @@
 import BraintreeDropIn
-import Braintree
+//import Braintree
```
sestevens, line 2:
```suggestion
```

Demo/Cartfile
```diff
@@ -0,0 +1,2 @@
+github "https://github.com/braintree/braintree_ios.git" "remove-spmtest"
```
sestevens, line 1: Currently, we're using a branch where we deleted the SPMTest demo app (since it causes Carthage to timeout if it's present). Once we do another release of the base SDK, we'll update the Cartfile to use that release. It'll pull the pre-built frameworks from the GitHub release instead of building them, so we won't have an issue with the SPMTest app.

.github/workflows/testing.yml
```diff
@@ -8,22 +8,26 @@ jobs:
...
-        run: pod install
+      - name: Use Xcode 12.3
+        run: sudo xcode-select -switch /Applications/Xcode_12.3.app
+      - name: Install xcpretty
+        run: gem install xcpretty
+      - name: Install Carthage dependencies
+        run: cd Demo && ../carthage.sh bootstrap
       - name: Run Unit Tests
-        run: set -o pipefail && xcodebuild -workspace 'BraintreeDropIn.xcworkspace' -sdk 'iphonesimulator' -configuration 'Release' -scheme 'UnitTests' -destination 'name=iPhone 11,platform=iOS Simulator'  test  | ./Pods/xcbeautify/xcbeautify
+        run: set -o pipefail && xcodebuild -workspace 'BraintreeDropIn.xcworkspace' -sdk 'iphonesimulator' -configuration 'Debug' -scheme 'UnitTests' -destination 'name=iPhone 11,platform=iOS Simulator'  test  | xcpretty
```
sshropshire, line 18: curious: is `Debug` more favorable for unit testing for the Drop-In project?

sshropshire (approved):

.github/workflows/testing.yml
scannillo: Good question. So running in `Debug` mode will build solely for the simulator architecture, whereas `Release` will build for all potential architectures (device & simulator).

But when running in `Release` mode we get the error: `building for iOS Simulator, but linking for architecture arm64`.

Since Xcode 12 introduces arm64 as a valid device architecture for Apple Silicon Macs, and it is also a valid simulator architecture - I think Xcode is getting confused on which one it is building for. Also, our Intel macs don't offer arm64 simulator support.

scannillo: It would have been nice if Apple named their new Apple Silicon architecture to arm64S or something. With XCode 12 we're having to do a handful of weird workarounds for this new "feature". So for now - running our tests in Debug works.

We confirmed everything runs fine on an actual device, too.

SampleApps/CarthageTest/Cartfile
```diff
@@ -1 +1 @@
-github "braintree/braintree-ios-drop-in"
\ No newline at end of file
+git "file:///Users/scannillo/bt/braintree-ios-drop-in" "spm-support"
```
sarahkoop, line 1: Will this be reverted to use the github repo in a later PR?

sarahkoop (approved):

SampleApps/CarthageTest/Cartfile
scannillo: Good catch - this looks like a remnant of our `rake` task when we were doing some local testing.

demerino (approved):

PR #258, opened 2021-01-13
Add Client SDK Deprecation Policy

scannillo:
### Summary of changes

 - Add Client SDK Deprecation policy link & chart to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,15 @@ Take a look at `BTCardFormViewController.m` to see examples of using the form fi
 
 Start with [**'Hello, Client!'**](https://developers.braintreepayments.com/ios/start/hello-client) for instructions on basic setup and usage.
 
+## Versions
+
+This SDK abides by our Client SDK Deprecation Policy. For more information on the potential statuses of an SDK check our [developer docs](http://developers.braintreepayments.com/guides/client-sdk/deprecation-policy).
+
+| Major version number | Status | Released | Deprecated | Unsupported |
+| -------------------- | ------ | -------- | ---------- | ----------- |
+| 9.x.x | Beta | March 2021 | TBA | TBA |
+| 8.x.x | Active | Februrary 2020 | TBA | TBA |
+
 ## Demo
 
 A demo app is included in project. To run it:
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ Welcome to Braintree's Drop-In SDK for iOS!
 
 **The Braintree iOS Drop-In SDK requires Xcode 12+**. It permits a deployment target of iOS 12.0 or higher.
 
+## Table of Contents
+
+1. [Getting Started](#getting-started)
+1. [Documentation](#documentation)
+1. [Demo App](#demo)
+1. [Help](#help)
+1. [Feedback](#feedback)
+
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -248,6 +248,15 @@ Take a look at `BTCardFormViewController.m` to see examples of using the form fi
 
 Start with [**'Hello, Client!'**](https://developers.braintreepayments.com/ios/start/hello-client) for instructions on basic setup and usage.
 
+## Versions
+
+This SDK abides by our Client SDK Deprecation Policy. For more information on the potential statuses of an SDK check our [developer docs](http://developers.braintreepayments.com/guides/client-sdk/deprecation-policy).
```
hollabaq86, line 253: abides

hollabaq86:
```suggestion
This SDK abides by our Client SDK Deprecation Policy. For more information on the statuses of an SDK, check our [developer docs](http://developers.braintreepayments.com/guides/client-sdk/deprecation-policy).[Minimum supported versions](https://developers.braintreepayments.com/guides/client-sdk/setup/ios/#requirements) are also available in our developer docs.
```

hollabaq86: One note about my suggested changes - these link to the minimum supported version of the Core SDK. I'm wondering if we should use the link to Drop-In's [setup page](https://developers.braintreepayments.com/guides/drop-in/setup-and-integration/ios/v5#setup) instead. Only think if the Drop-In setup page doesn't mention minimum supported versions... maybe we should update that.

hollabaq86 (commented):
Some suggested tweaks to the language to bring it inline with the server SDK blurbs about deprecation policies.

One other general note - the README for this library is getting pretty large, what do folks think about adding a Table of Contents with links to the top?

README.md
scannillo: Good call - I added a JIRA (#DTBTSDK-720) to update the Drop-In Developer Docs with the minimum supported versions. 

Once we do - we can make this change to the README. How does that sound, and we add that in a later PR?

README.md
```diff
@@ -248,6 +248,15 @@ Take a look at `BTCardFormViewController.m` to see examples of using the form fi
...
+
+| Major version number | Status | Released | Deprecated | Unsupported |
+| -------------------- | ------ | -------- | ---------- | ----------- |
+| 9.x.x | Beta | March 2021 | TBA | TBA |
```
sestevens, line 257:
```suggestion
| 9.x.x | Beta | TBA | TBA | TBA |
```

Should we just put TBA for the v9 release date? Or omit it for now and add it when we actually do a beta release?

sestevens:
+1 to @hollabaq86's idea of adding a table of contents to the README.

README.md
```diff
@@ -12,6 +12,14 @@ Welcome to Braintree's Drop-In SDK for iOS!
 
 **The Braintree iOS Drop-In SDK requires Xcode 12+**. It permits a deployment target of iOS 12.0 or higher.
 
+## Table of Contents
+
+1. [Getting Started](#getting-started)
+1. [Documentation](#documentation)
+1. [Demo App](#demo)
+1. [Help](#help)
+1. [Feedback](#feedback)
```
sestevens, line 21: Should we add the Versions and License sections here?

sestevens (approved):
Looks good to me! Good call adding the TODO for v9. That would be easy to forget.

PR #261, opened 2021-01-22
Replace BTTokenizationService with BTVenmoDriver

scannillo:
### Summary of changes
 - Replace `BTTokenizationService` with `BTVenmoDriver` in `BTPaymentSelectionViewController`
 - Make header files that didn't need to be public -> private & update import syntax

 ### Checklist
 - [X] Added a changelog entry

### Authors
@sestevens @scannillo


UnitTests/MockAPIClient.swift
```diff
@@ -0,0 +1,3 @@
+@objc class MockAPIClient: BTAPIClient {
```
sestevens, line 1: Right now this doesn't really do anything, but I think in the future it will be helpful to have a `MockAPIClient` in Drop-in.

Sources/BraintreeDropIn/BTCardFormViewController.h
```diff
@@ -1,4 +1,4 @@
-#import <BraintreeDropIn/BTDropInBaseViewController.h>
+#import "BTDropInBaseViewController.h"
```
sestevens, line 1: These import syntax changes are the result of making some public headers private (since merchants shouldn't be using them directly anyway).

sshropshire (approved):
🌊

sarahkoop (approved):

demerino (approved):

PR #264, opened 2021-01-22
Update header imports

scannillo:
### Summary of changes

 - Update DEVELOPMENT.md with instructions for import syntax used to satisfy all dependency managers
 - Update imports throughout SDK to use COCOAPODS macro and < > (modular) imports

 ### Checklist

 - ~Added a changelog entry~

### Authors
@sestevens @scannillo

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,63 +1,61 @@
-# Braintree-iOS Development Notes
+# Braintree iOS Drop-in Development Notes
 
-This document outlines development practices that we follow internally while developing this SDK.
+This document outlines development practices that we follow while developing this SDK.
 
 ## Development Merchant Server
 
-The included demo app utilizes a test merchant server hosted on heroku ([https://braintree-sample-merchant.herokuapp.com](https://braintree-sample-merchant.herokuapp.com)). It
-produces client tokens that point to Braintree's Sandbox Environment.
+The included demo app utilizes a [sandbox sample merchant server](https://braintree-sample-merchant.herokuapp.com) hosted on Heroku.
 
 ## Tests
 
-There are a number of test targets for each section of the project. You can run all tests on the command line with `bundle && rake spec:all`. 
+There are a number of test targets for each section of the project.
 
 It's a good idea to run `rake`, which runs all unit tests, before committing.
 
-The integration tests require a full Braintree stack running on localhost.
+Use the following commands to run tests:
+* UI tests: `bundle && rake spec:ui`
+* Unit tests: `bundle && rake spec:unit`
+* All tests: `bundle && rake spec:all`
 
-## Architecture
+## Importing Header Files
 
-See [Frameworks](Frameworks.markdown) for an overview of the components that comprise this SDK.
+To maintain support for CocoaPods, Swift Package Manager, and Carthage, our Objective-C import statements need specific attention.
 
-## Environmental Assumptions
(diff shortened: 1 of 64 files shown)


Sources/BraintreeDropIn/BTCardFormViewController.h
```diff
@@ -1,5 +1,6 @@
 #import "BTDropInBaseViewController.h"
-#if __has_include(<BraintreeDropIn/BraintreeUIKit.h>)
+
```
scannillo, line 2:
```suggestion
```

Sources/BraintreeDropIn/BTDropInUIUtilities.h
```diff
@@ -1,5 +1,5 @@
 #import <UIKit/UIKit.h>
-#if __has_include(<BraintreeDropIn/BraintreeUIKit.h>)
+#ifdef COCOAPODS
```
scannillo, line 2:
```suggestion

#ifdef COCOAPODS
```

Sources/BraintreeDropIn/Custom Views/BTDropInPaymentSeletionCell.h
```diff
@@ -1,5 +1,5 @@
 #import <UIKit/UIKit.h>
-#if __has_include(<BraintreeDropIn/BraintreeUIKit.h>)
+#ifdef COCOAPODS
```
scannillo, line 2:
```suggestion

#ifdef COCOAPODS
```

DEVELOPMENT.md
```diff
@@ -1,63 +1,61 @@
...
+#else // SPM or Carthage
+#import <BraintreeUIKit/BraintreeUIKit.h>
+#endif
+```
 
-The release process is self-documented in a number of rake tasks.
+To import a Braintree framework written in **Swift** into an Objective-C file, use the following syntax:
+```objc
+#ifdef COCOAPODS // CocoaPods
+#import <BraintreeDropIn/BraintreeDropIn-Swift.h>
```
sarahkoop, line 53: Should this import filename be consistent with the `YourNewSwiftFramework` example naming below?

sshropshire, line 53: question: wouldn't this be `<YourNewSwiftFramework/YourNewSwiftFramework-Swift.h>` as well?

sarahkoop (approved):

sshropshire (approved):

DEVELOPMENT.md
scannillo: So - this was actually meant to be `#import <BraintreeDropIn/BraintreeDropIn-Swift.h>`. Since we specified the source code files in our Podspec, CocoaPods will automatically create a `<Module>-Swift.h` umbrella header.

An example of importing a Swift framework into an Objective-C file is the `PayPalDataCollector` in braintree_ios. [In this example, ](https://github.com/braintree/braintree_ios/blob/9cdff39ba8155efe38f47ceabb8a05e585991a69/Sources/BraintreePayPal/BTPayPalDriver.m#L6-L33)the last line of each import group represents the import syntax for including. PayPalDataCollector.

sshropshire: Aah got it. Does the merchant ever use `@import Braintree;`? I'm SPM noobing right now lol.

scannillo: Yup, for merchants using Objective-C, they can use the `@import` syntax if they want to. The `@ import` is the newer, "module" import syntax. It's more efficient that `#import` in that it handles automatically including the framework in your project, is compiled only once, and loads a binary representation of the framework (faster compile times).

PR #270, opened 2021-02-22
Add BTDropInUICustomization

sestevens:
### Summary of changes

 - Rather than setting properties on `BTUIKAppearance` directly, allow merchants to customize Drop-in's UI by creating a `BTDropInUICustomization` and setting the `uiCustomization` property on `BTDropInRequest`
 - If `uiCustomization` is nil, use the light color scheme as a default

 ### Checklist

 - [x] Added a changelog entry

- @sestevens


Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInUICustomization.h
```diff
@@ -0,0 +1,98 @@
...
++ (instancetype)new __attribute__((unavailable("Please use initWithColorScheme:")));
+
+- (instancetype)init __attribute__((unavailable("Please use initWithColorScheme:")));
+
+/// Creates an instance of BTDropInUICustomization with the given color scheme.
+///
+/// @param colorScheme The color scheme used to determine the default colors used. These defaults may be overriden by setting additional properties on BTDropInUICustomization.
+///
+/// @return UI customization options for Braintree Drop-in
+- (instancetype)initWithColorScheme:(BTDropInColorScheme)colorScheme;
```
sestevens, line 29: My thought with this initializer is that default values will be set based on the color scheme when the `BTDropInUICustomization` is initialized. And then merchants can override specific properties as needed.

The way things currently work with `BTUIKAppearance`, you get different results based on whether you set the color scheme first and _then_ set other properties or vice versa. (Basically, whatever you do last gets priority, but it's a little weird that the order of operations matters.)

scannillo:
This looks awesome!

Question - can we make `BTUIKAppearance.h` non public now?

sestevens:
> This looks awesome!
> 
> Question - can we make `BTUIKAppearance.h` non public now?

I was thinking about this too! Currently, I think it needs to be public because it lives in `BraintreeUIKit` and needs to be accessed from `BraintreeDropIn`. We could consider merging the two frameworks into one, though. That might give us more freedom to make code changes to the classes in `BraintreeUIKit` and limit what we make public.

sshropshire (approved):
🌊

Sources/BraintreeDropIn/BTDropInUICustomization.m
```diff
@@ -0,0 +1,79 @@
+#import "BTDropInUICustomization.h"
+
+#ifdef COCOAPODS
+#import <BraintreeDropIn/UIColor+BTUIK.h>
+#else
+#import <BraintreeUIKit/UIColor+BTUIK.h>
+#endif
+
+@implementation BTDropInUICustomization
```
sshropshire, line 9: Also agreed this is dope! Would there be a way for merchants to supply a custom color scheme in the future? I think it's possible on Android with xml themes, but would have to double-check. Merchants could potentially supply their own color schemes for both light and dark mode.

sestevens: Good question! My thought was that if merchants want to create their own color schemes, they would first specify one of our color schemes (light, dark or dynamic) and then customize additional properties as needed. Theoretically, they could end up overriding every property, so it wouldn't actually matter which color scheme they passed into the initializer.

demerino (approved):
This looks great, much more developer friendly.

scannillo (approved):
🚀

PR #271, opened 2021-02-23
Add support for dynamic type

sestevens:
### Summary of changes

 - Enable dynamic type by default, for both system and custom fonts
 - Add `disableDynamicType` property to `BTDropInUICustomization`
 - Adjust layout constraints as needed to allow font to scale

Work for follow-up PRs:
- Adjust auto layout constraints on `BTEnrollmentVerificationViewController`, which is used for Union Pay.
- On payment selection screen, make the vaulted payment methods collection view scale vertically (currently, labels are clipped in very large fonts).
- On card form, stack each label and text field vertically instead of horizontally for large font sizes. (Similar to the label and switch at the end of [this post](https://useyourloaf.com/blog/making-space-for-dynamic-type/).)
- On card number input page, dismiss the keyboard if the user taps `Next` and an error occurs. Currently, the keyboard covers the error message when the font is large.

It would be great if a few folks could pull down this branch and play around with different font sizes. On the simulator, you can change the font size by going to Settings > Accessibility > Display & Text Size > Larger Text. Or you can use the [Environment Overrides](https://useyourloaf.com/blog/xcode-11-environmental-overrides/) option in Xcode.

❓  **QUESTION:** We currently display "••• ••11" underneath vaulted cards in the payment selection sheet, but that truncates when the font is too big. I'm thinking we should just display the last four digits of the card number instead. What do people think of that idea?

 ### Checklist

 - [x] Added a changelog entry
 - [x] Updated migration guide

- @sestevens 
- @scannillo 
- @hollabaq86 

### Screenshots

Payment selection screen:

Simulator Screen Shot - iPhone 12 mini - 2021-02-25 at 13 42 27

--------------

Card form (1):

Simulator Screen Shot - iPhone 12 mini - 2021-02-25 at 13 42 38

--------------

Card form (2):

Simulator Screen Shot - iPhone 12 mini - 2021-02-25 at 13 43 04

**NOTE:** The expiration date is definitely too small here. The plan is to have the text fields appear below the labels instead of side by side when the font size is large.


UnitTests/BTDropInUICustomizationTests.swift
```diff
@@ -86,6 +88,61 @@ class BTDropInUICustomizationTests: XCTestCase {
             XCTAssertNil(uiCustomization.fontFamily)
             XCTAssertNil(uiCustomization.boldFontFamily)
             XCTAssertNil(uiCustomization.navigationBarTitleTextColor)
+            XCTAssertFalse(uiCustomization.disableDynamicType)
         }
     }
+
+    func testCustomFonts_withDynamicTypeDisabled() {
```
sestevens, line 95: I think these tests should actually go in `BTUIKAppearance+DropInTests.swift` since they're testing the `configure` method on `BTUIKAppearance+DropIn`.

Sources/BraintreeDropIn/Custom Views/BTUIPaymentMethodCollectionViewCell.m
```diff
@@ -35,7 +35,7 @@ - (instancetype)initWithFrame:(CGRect)frame
         [self.contentView addSubview:self.paymentOptionCardView];
 
         self.titleLabel = [[UILabel alloc] init];
-        [BTUIKAppearance styleSmallLabelBoldPrimary:self.titleLabel];
+        [BTUIKAppearance styleSmallLabelPrimary:self.titleLabel];
```
sestevens, line 38: This was the only place that the `styleSmallLabelBoldPrimary` method was used, and Apple doesn't provide a bold dynamic font style that small enough to match what we had previously. Opted to make it not bold instead. (This is the title -- e.g., PayPal, Visa, MasterCard -- that appears under the icon in the vaulted payment method collection view.)

Sources/BraintreeUIKit/Components/BTUIKCardNumberFormField.m
```diff
@@ -50,18 +50,12 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
-        [self.validateButton setAttributedTitle:normalValidateButtonString
-                                       forState:UIControlStateNormal];
-        NSAttributedString *disabledValidateButtonString = [[NSAttributedString alloc] initWithString:BTUIKLocalizedString(NEXT_ACTION)
-                                                                                           attributes:@{NSForegroundColorAttributeName:[BTUIKAppearance sharedInstance].disabledColor,
-                                                                                                        NSFontAttributeName:[[BTUIKAppearance sharedInstance].boldFont fontWithSize:UIFont.labelFontSize]}];
-        [self.validateButton setAttributedTitle:disabledValidateButtonString forState:UIControlStateDisabled];
+        [self.validateButton setTitleColor:[BTUIKAppearance sharedInstance].tintColor forState:UIControlStateNormal];
+        [self.validateButton setTitleColor:[BTUIKAppearance sharedInstance].disabledColor forState:UIControlStateDisabled];
 
+        self.validateButton.titleLabel.font = [BTUIKAppearance sharedInstance].staticHeadlineFont;
```
sestevens, line 58: Using an attributed title didn't seem necessary here, so we decided to simplify the code a bit.

Sources/BraintreeUIKit/Components/BTUIKFormField.m
```diff
@@ -78,26 +79,20 @@ - (void)updateConstraints {
...
     BOOL hasFormLabel = (self.formLabel.text.length > 0);
     
-    [self.layoutConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:|[textField]|"
+    [self.layoutConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:|-(PADDING)-[textField]-(PADDING)-|"
                                                                  options:0
                                                                  metrics:metrics
                                                                    views:viewBindings]];
 
-    [self.layoutConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:|[formLabel]|"
+    [self.layoutConstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:|-(PADDING)-[formLabel]-(PADDING)-|"
```
sestevens, line 87: Since we can't use a fixed height for these form fields anymore, it made sense to add in a little padding above and below so that things don't look too squished.

Sources/BraintreeUIKit/Helpers/UIFont+BTUIK.m
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            return [[UIFontMetrics metricsForTextStyle:UIFontTextStyleBody] scaledFontForFont:customFont];
+        }
+    } else {
+        if (useStaticSize) {
+            return [UIFont systemFontOfSize:UIFont.labelFontSize];
+        } else {
+            return [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
+        }
+    }
```
sestevens, line 23: There are four cases for each font:
- Custom font with dynamic type disabled
- Custom font with dynamic type enabled
- System font with dynamic type disabled
- System font with dynamic type enabled

sarahkoop (approved):
Looks good to me! I tested out a few different font sizes on the simulator and they looked good. I agree that the card number truncation would look better for large font sizes if it just showed 4 numbers rather than the dots.

sshropshire (approved):

PR #273, opened 2021-02-26
Add Rakefile tasks for publishing Drop-in reference docs

sestevens:
### Summary of changes

 - Add Rakefile tasks for publishing Drop-in reference docs
 - Update README so that we don't have broken links in the reference docs or on CocoaPods

You can checkout the generated docs [here](https://braintree.github.io/braintree-ios-drop-in/). I will also add a link on our [dev docs](https://developers.braintreepayments.com/reference/overview). Eventually, we can add a GitHub actions release workflow with a step to generate these docs.

 ### Checklist

 - ~Added a changelog entry~

- @sestevens


Rakefile
```diff
@@ -227,3 +227,46 @@ namespace :publish do
...
   end
 end
+
+def jazzy_command
+  %W[jazzy
+      --objc
+      --author Braintree
+      --author_url http://braintreepayments.com
+      --github_url https://github.com/braintree/braintree-ios-drop-in
+      --github-file-prefix https://github.com/braintree/braintree-ios-drop-in/tree/8.1.2
```
sestevens, line 237:
```suggestion
      --github-file-prefix https://github.com/braintree/braintree-ios-drop-in/tree/#{current_version}
```

Rakefile
```diff
@@ -227,3 +227,39 @@ namespace :publish do
...
+  desc "Generate docs with jazzy"
+  task :generate do
+    run(jazzy_command)
+    run! "cp -R Images #{current_version}/Images" # copy images used in README
+    puts "Generated HTML documentation"
+  end
+
+  task :publish do
+    run! "git checkout gh-pages"
+    run! "ln -sfn #{current_version} current" # update symlink to current version
```
sestevens, line 258: I structured things like the [docs for the JS SDK](https://github.com/braintree/braintree-web/tree/gh-pages) with a folder for each SDK version and a `current` directory that is just a symlink to the folder for the most recent version.

sshropshire (approved):

scannillo (approved):
YAY! Thank you for doing this. We finally have iOS drop in reference docs now 😎

PR #275, opened 2021-03-01
Add release workflow using GitHub Actions

sestevens:
### Summary of changes

 - Add release workflow using GitHub Actions
 - Fix warnings in Podspec related to invalid screenshot URL

 ### Checklist

 - ~Added a changelog entry~

- @sestevens 
- @scannillo

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,15 +14,19 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+
       - name: Use Xcode 12.4
         run: sudo xcode-select -switch /Applications/Xcode_12.4.app
 
-      # Check for unreleased section in changelog
       - name: Check for unreleased section in changelog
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+    runs-on: macOS-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Use Xcode 12.4
+        run: sudo xcode-select -switch /Applications/Xcode_12.4.app
+
+      # Check for unreleased section in changelog
```
sarahkoop, line 20: Nitpick: some of these comments seem helpful, but others are identical to the names - maybe we could make all the names explanatory and remove the extra comments?

sarahkoop (approved):

sshropshire (approved):
🌊

.github/workflows/release.yml
sestevens: Good call -- we updated the names and removed the comments.

PR #277, opened 2021-03-01
Show last four digits of vaulted cards

sestevens:
### Summary of changes

* Show last four digits of card number instead of "••• ••11". This prevents the important info (the digits) from truncating when a large font size is used.

This change applies to:
* Payment method selection screen (when vaulted payment methods are shown)
* Vault manager screen

 ### Checklist

 - [x] Added a changelog entry

### Authors
-@sestevens

### Screenshots

*Payment method selection screen:*

Simulator Screen Shot - iPhone 12 - 2021-03-01 at 17 17 42

----------------

*Vault manager screen:*

Simulator Screen Shot - iPhone 12 - 2021-03-01 at 17 17 49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     * `BTVaultManagementViewController`
   * Add `BTDropInUICustomization`, which replaces `BTUIKAppearance` for customizing UI
   * Enable dynamic type by default. This can be disabled with the `disableDynamicType` property on `BTDropInUICustomization`.
+  * Show last four digits of vaulted cards, instead of "••• ••11", which truncates in larger fonts.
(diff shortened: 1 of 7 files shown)


scannillo (approved):
I like it!

scannillo:
We should also update the screenshot in the README for v9, since it includes the "··· ··11" style.

sarahkoop (approved):
Looks good 🌟

sestevens:
> We should also update the screenshot in the README for v9, since it includes the "··· ··11" style.

Good call! I'll make a note in our v9 doc. We can create new screenshots before we release, in case there are any other UI changes before then.

PR #281, opened 2021-03-05
Update demo app to include all supported localizations

sestevens:
### Summary of changes

 - Update the demo app so that we can easily test any supported language. (The demo app itself will still be in English, but we'll be able to easily change the device's language to see how Drop-in looks in any of our supported languages.)
 - Update demo app to use stack views so that the UI doesn't break when a right-to-left language is selected.

 ### Checklist

 - ~Added a changelog entry~

- @sestevens

### DIFF
--- a/Demo/Supporting Files/ar.lproj/InfoPlist.strings
+++ b/Demo/Supporting Files/ar.lproj/InfoPlist.strings
@@ -1,0 +1,2 @@
+/* Localized versions of Info.plist keys */
+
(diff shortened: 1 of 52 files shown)


Demo/Supporting Files/ar.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,2 @@
+/* Localized versions of Info.plist keys */
```
sestevens, line 1: No need to review any of these `.lproj` files -- everything from here down is auto-generated.

sarahkoop (approved):

sshropshire (approved):

PR #282, opened 2021-03-08
Union pay dynamic fonts

sestevens:
### Summary of changes

 - Update braintree_ios to version 5.1.0 in Cartfile.resolved. (The Cartfile itself specifies ~> 5.0.)
 - Allow text field on Union Pay SMS enrollment screen to scale vertically
 - Maintain horizontal padding around SMS sent label with larger fonts
 - Update "Use a different phone number button" to match the style of other buttons (i.e. bold and non-scaling)

 ### Checklist

 - ~Added a changelog entry~ (We already have an entry for dynamic type)

- @sestevens

### Screenshots

Simulator Screen Shot - iPhone 12 - 2021-03-08 at 16 12 40

-------------------

Simulator Screen Shot - iPhone 12 - 2021-03-08 at 16 13 14

---------------------

Simulator Screen Shot - iPhone 12 - 2021-03-08 at 16 13 52

--------------

Simulator Screen Shot - iPhone 12 - 2021-03-08 at 16 13 44

### DIFF
--- a/Sources/BraintreeDropIn/Custom Views/BTEnrollmentVerificationViewController.m
+++ b/Sources/BraintreeDropIn/Custom Views/BTEnrollmentVerificationViewController.m
@@ -69,16 +69,12 @@
     self.resendSmsButton = [UIButton new];
     self.resendSmsButton.translatesAutoresizingMaskIntoConstraints = NO;
     [self.resendSmsButton setTitle:smsButtonText forState:UIControlStateNormal];
-    
-    NSAttributedString *normalValidateButtonString = [[NSAttributedString alloc] initWithString:smsButtonText
-                                                                                     attributes:@{NSForegroundColorAttributeName:[BTUIKAppearance sharedInstance].tintColor,
-                                                                                                  NSFontAttributeName:[BTUIKAppearance sharedInstance].bodyFont}];
-    [self.resendSmsButton setAttributedTitle:normalValidateButtonString forState:UIControlStateNormal];
-    NSAttributedString *disabledValidateButtonString = [[NSAttributedString alloc] initWithString:smsButtonText
-                                                                                       attributes:@{NSForegroundColorAttributeName:[BTUIKAppearance sharedInstance].disabledColor,
-                                                                                                    NSFontAttributeName:[BTUIKAppearance sharedInstance].bodyFont}];
-    [self.resendSmsButton setAttributedTitle:disabledValidateButtonString forState:UIControlStateDisabled];
-    
+
+    [self.resendSmsButton setTitleColor:[BTUIKAppearance sharedInstance].tintColor forState:UIControlStateNormal];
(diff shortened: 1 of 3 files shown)


Sources/BraintreeUIKit/Components/BTUIKFormField.m
```diff
@@ -36,6 +36,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
         self.textField.adjustsFontSizeToFitWidth = YES;
         self.textField.returnKeyType = UIReturnKeyNext;
         self.textField.keyboardAppearance = [BTUIKAppearance sharedInstance].keyboardAppearance;
+        self.textField.textAlignment = [BTUIKViewUtil naturalTextAlignment];
```
sestevens, line 39: I changed this to default to natural text alignment, which is what we always want if the text field doesn't have a label. When label text is set, there's logic to determine the correct text alignment for the text field, depending on the font size.

Sources/BraintreeDropIn/Custom Views/BTEnrollmentVerificationViewController.m
```diff
@@ -65,44 +69,33 @@ - (void)viewDidLoad {
...
-    NSAttributedString *disabledValidateButtonString = [[NSAttributedString alloc] initWithString:smsButtonText
-                                                                                       attributes:@{NSForegroundColorAttributeName:[BTUIKAppearance sharedInstance].disabledColor,
-                                                                                                    NSFontAttributeName:[BTUIKAppearance sharedInstance].bodyFont}];
-    [self.resendSmsButton setAttributedTitle:disabledValidateButtonString forState:UIControlStateDisabled];
-    
+
+    [self.resendSmsButton setTitleColor:[BTUIKAppearance sharedInstance].tintColor forState:UIControlStateNormal];
+    [self.resendSmsButton setTitleColor:[BTUIKAppearance sharedInstance].disabledColor forState:UIControlStateDisabled];
+
+    self.resendSmsButton.titleLabel.font = [BTUIKAppearance sharedInstance].staticHeadlineFont;
```
sarahkoop, line 76: Can you remind me why this button font doesn't scale along with dynamic text?

sshropshire (approved):

sarahkoop (approved):

Sources/BraintreeDropIn/Custom Views/BTEnrollmentVerificationViewController.m
sestevens: We've been giving the buttons a static font size throughout Drop-in because that matches what Apple seems to do with buttons. I'm not sure if this is a place where we'd want to break that rule, though, since we do have more space on the page? We can also make a note to run this by the accessibility team and see what they think.

sarahkoop: Yeah following what Apple does makes sense to me. Running it by the accessibility team would probably be a good idea, though.

PR #283, opened 2021-03-09
Update Demo & BraintreeDropIn target to use xcframework dependencies

scannillo:
### Background

 - Carthage 0.37.0 offers support for building dependencies as `xcframeworks`
 - Both our Demo app and the BraintreeDropIn target use Carthage to fetch Braintree (& InAppSettingsKit) dependencies

### Changes

- Update Demo app and the BraintreeDropIn target to use `xcframework` dependencies
- Remove run script phases
- Update README with instructions

 ### Checklist

 - ~Added a changelog entry~

### Authors
@hollabaq86 @scannillo


.github/workflows/tests.yml
```diff
@@ -20,10 +20,10 @@ jobs:
         uses: actions/cache@v2
         with:
           path: Carthage
-          key: ${{ runner.os }}-carthage-${{ hashFiles('**/Cartfile.resolved') }}
+          key: ${{ runner.os }}-carthage-${{ hashFiles('**/Cartfile.resolved') }}-3
```
sestevens, line 23: Apparently there's [no way to clear the cache](https://github.com/actions/cache/issues/2), so I just updated the key. The cache had regular frameworks instead of xcframeworks.

sestevens (approved):

demerino:
What would be required for us to drop this Carthage requirement for the demo app?

scannillo:
> What would be required for us to drop this Carthage requirement for the demo app?

We plan to revisit this in a future PR. A few weeks ago, Susan and I had trouble with the new project structure using CocoaPods or SPM for the demo. Carthage was what we got working, and since it unblocked development for the rest of the SDK, we ran with it.

sestevens:
I spiked out using SPM to add the base SDK dependencies to the Drop-in framework, and it seems to work. The demo app still uses the Drop-in framework directly (not through any package manager). The branch is [spm-demo](https://github.com/braintree/braintree-ios-drop-in/compare/next-major-version...spm-demo). I tested `pod lib lint` and `rake spm:build_demo`, and both ran successfully. I still couldn't get the Carthage test app to work, though.

scannillo:
I'm gonna merge this - since our demo app isn't really working without it. We can make Susan's SPM changes in a future PR!

PR #286, opened 2021-03-16
Fix crash on slow network connections

sestevens:
### Summary of changes

- Resolve crash reported in #243 

The crash occurred after entering the card number if the configuration had not finished loading. Drop-in tried to auto-advance to the next field, but if the configuration hadn't finished loading, we didn't have the list of required fields yet, so there was nothing to advance to. The fix for this is to check if the index of the current field in the list of required fields is `NSNotFound`, and if so, skip the attempt to advance to the next field.

Additionally, we were copying `BTAPIClient` for each view controller that subclasses `BTDropInBaseViewController`. This meant that each view controller had to fetch the configuration, instead of using the cached configuration previously fetched. If we use the same instance of `BTAPIClient` across view controllers, we'll reduce latency by using the cached configuration and avoid any race conditions (like the one above).

 ### Checklist

 - [x] Added a changelog entry

- @sestevens

### DIFF
--- a/BraintreeDropIn/BTCardFormViewController.m
+++ b/BraintreeDropIn/BTCardFormViewController.m
@@ -143,7 +143,7 @@
     [self resetForm];
     [self showLoadingScreen:YES];
     [self loadConfiguration];
-    
+
     self.firstResponderFormField = self.cardNumberField;
 }
 
@@ -472,7 +472,7 @@
 
 - (void)advanceFocusFromField:(BTUIKFormField *)currentField {
     NSUInteger currentIdx = [self.requiredFields indexOfObject:currentField];
-    if (currentIdx < [self.requiredFields count] - 1) {
+    if (currentIdx != NSNotFound && currentIdx < self.requiredFields.count - 1) {
         [[self.requiredFields objectAtIndex:currentIdx + 1] becomeFirstResponder];
     }
 }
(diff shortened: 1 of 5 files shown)


BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -472,7 +472,7 @@ - (void)updateSubmitButton {
 
 - (void)advanceFocusFromField:(BTUIKFormField *)currentField {
     NSUInteger currentIdx = [self.requiredFields indexOfObject:currentField];
-    if (currentIdx < [self.requiredFields count] - 1) {
+    if (currentIdx != NSNotFound && currentIdx < self.requiredFields.count - 1) {
         [[self.requiredFields objectAtIndex:currentIdx + 1] becomeFirstResponder];
```
sestevens, line 476: The crash was occurring on this line. If the configuration hasn't finished loading, `self.requiredFields` will be an empty array, which means we'll crash when we try to access the object at index `currentIdx +1`.

BraintreeDropIn/BTCardFormViewController.m
```diff
@@ -682,30 +682,32 @@ - (void)enrollCard {
...
+
+    if (!formField.valid) {
+        [self cardNumberErrorHidden:NO];
+        return;
+    }
+
+    if (!self.configuration) {
+        self.cardNumberField.state = BTUIKCardNumberFormFieldStateLoading;
+        return;
+    }
```
sestevens, line 701: This is the main addition to this method -- if we don't yet have the configuration, show the loading indicator next to the card number form instead of trying to advance to the next field.

BraintreeDropIn/BTDropInBaseViewController.m
```diff
@@ -19,7 +19,7 @@ @implementation BTDropInBaseViewController
 - (instancetype)initWithAPIClient:(BTAPIClient *)apiClient request:(BTDropInRequest *)request
 {
     if (self = [super init]) {
-        self.apiClient = [apiClient copyWithSource:apiClient.metadata.source integration:BTClientMetadataIntegrationDropIn2];
+        self.apiClient = apiClient;
```
sestevens, line 22: This lets us access the cached configuration fetched previously, instead of having to re-fetch it for each view controller.

scannillo: I think this will change the [metadata](https://github.com/braintree/braintree_ios/blob/2946eebfdef0729dade1f7994ff1869145a04a1d/Sources/BraintreeCore/BTClientMetadata.m#L61-L72) sent in Drop-In's analytics requests. We send `self.apiClient.metadata.integrationString` throughout the Base SDK.

BraintreeDropIn/Public/BTCardFormViewController.h
```diff
@@ -43,7 +43,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable, readonly) BTCardCapabilities *cardCapabilities;
 
 /// The card network types supported by this merchant
-@property (nonatomic, copy) NSArray *supportedCardTypes;
+@property (nonatomic, copy) NSArray<NSNumber *> *supportedCardTypes;
```
scannillo, line 46: Can we use `<BTUIKPaymentOptionType *>` with the enum name here to make this more readable?

sestevens: Unfortunately, we can't. In Objective-C, NSArrays can only hold NSObjects, and `BTUIKPaymentOptionType` is an enum, not an NSObject.

BraintreeDropIn/BTDropInBaseViewController.m
sestevens: The metadata is being set on the api client when it's initially passed into the `BTDropInController`, [here](https://github.com/braintree/braintree-ios-drop-in/blob/4ece967437cf6db3db299d10c0f3120d44391d93/BraintreeDropIn/BTDropInController.m#L61). Once it's been set once, we shouldn't need to continue setting it.

demerino (approved):

PR #287, opened 2021-03-17
Auto-scroll card number error view on screen

sestevens:
### Summary of changes

 - If the user is using larger font sizes, the card number error view may be hidden by the keyboard. This PR adds logic to automatically scroll so that the error message is visible.

 ### Checklist

 - ~[] Added a changelog entry~ (I think we can consider this part of the dynamic fonts updates)

- @sestevens


demerino (approved):
We did manual testing for the `regular` and `enlarged` fonts?

sestevens:
> We did manual testing for the `regular` and `enlarged` fonts?

Yep!

sshropshire (approved):

PR #289, opened 2021-03-19
Add Localization Support for SPM Integrations

sestevens:
### Summary of changes

 - Add support for localization with Swift Package Manager:
     - Move localized strings under a "Resources" directory (per [Apple's requirements](https://developer.apple.com/documentation/swift_packages/localizing_package_resources))
     - Update logic to get the bundle with the localized strings (see [this article](https://darjeelingsteve.com/articles/How-to-Use-Module-Resources-in-Objective-C-SPM-Packages.html))
 - Update the SPMTest app to include a simple UI test that checks that localizations are working
 - Update the `build` GitHub Actions workflow to run that UI test

I think we should add a similar UI test for CocoaPods and perhaps our demo app, but we can do that in a separate PR.

 ### Checklist

 - ~Added a changelog entry~ (this falls under the "Add support for SPM" entry)

- @sestevens


.github/workflows/build.yml
```diff
@@ -19,9 +19,13 @@ jobs:
         uses: actions/checkout@v2
       - name: Use Xcode 12.4
         run: sudo xcode-select -switch /Applications/Xcode_12.4.app
+      - name: Install xcpretty
```
scannillo, line 22: Random - when I was trialing testflight stuff via GH Actions, I didn't need to install `xcpretty` to use it. 🤷‍♀️

Sources/BraintreeUIKit/Localization/BTUIKLocalizedString.m
```diff
@@ -19,15 +19,18 @@ + (NSBundle *)localizationBundle {
...
-
-    // CocoaPods creates "Braintree-UIKit-Localization.bundle", so we check for that first.
-    // If not found, use the BraintreeUIKit bundle, which is where Carthage will find the localized strings.
-    return localizationBundlePath ? [NSBundle bundleWithPath:localizationBundlePath] : [NSBundle bundleForClass:self.class];
+    return [NSBundle bundleWithPath:localizationBundlePath];
+#elif SWIFT_PACKAGE
+    return SWIFTPM_MODULE_BUNDLE;
+#else
+    return [NSBundle bundleForClass:self.class];
+#endif
```
scannillo, line 33: Nice cleanup here. Also nice job researching this!

scannillo (approved):
🏁

.github/workflows/build.yml
sestevens: Oooh, good to know! I'll try removing it.

Sources/BraintreeUIKit/Localization/BTUIKLocalizedString.m
sestevens: Thank you! 😊

.github/workflows/build.yml
```diff
@@ -19,9 +19,13 @@ jobs:
...
+        run: gem install xcpretty
```
sestevens, line 23:
```suggestion
```

sarahkoop (approved):

PR #296, opened 2021-03-25
Use SPM for Development

sestevens:
### Summary of changes

 - Use Swift Package Manager to pull in dependencies for local development. The demo app and unit tests use a local Swift Package version of `BraintreeDropIn`. The demo app also uses the Swift Package for InAppSettingsKit.
 - Previously, we were using Carthage to pull in dependencies, so this PR also removes code related to Carthage.

 ### Checklist

- @sestevens 
- @scannillo


sshropshire (approved):

sarahkoop (approved):
🌟

PR #310, opened 2021-04-02
Remove `threeDSecureVerification` property

sestevens:
### Summary of changes

 - Remove `threeDSecureVerification` from `BTDropInRequest`; instead of setting this, merchants should set `threeDSecureRequest` in order to request 3DS

 ### Checklist

 - [x] Added a changelog entry

- @sestevens


sshropshire (approved):
🌊

Sources/BraintreeDropIn/BTDropInController.m
```diff
@@ -268,9 +268,6 @@ - (void)threeDSecureVerification:(BTPaymentMethodNonce *)tokenizedCard {
     paymentFlowDriver.viewControllerPresentingDelegate = self;
 
     BTThreeDSecureRequest *request = self.dropInRequest.threeDSecureRequest;
-    if (!request) {
-        request = [[BTThreeDSecureRequest alloc] init];
-    }
```
scannillo, line 273: I was thinking - this PR should also remove wherever the logic was that set the amount on the `ThreeDSecureRequest` from the `BTDropInResult` ... however after searching the SDK I don't see this happening anywhere. So this looks sufficient, then!

scannillo (approved):
YEE HAW

Sources/BraintreeDropIn/BTDropInController.m
sestevens: Yeah, I was looking for that too! I think we must have removed `amount` previously.

PR #324, opened 2021-04-23
Add SwiftUI demo to demo app

sestevens:
### Summary of changes

 - Add SwiftUI demo to the demo app. You can access it selecting "SwiftUI" from the "UI Frameworks" option in the Settings page.

This PR doesn't fix any issues with SwiftUI, and it's not really a fully fleshed out example of how to integrate with Drop-in using SwiftUI. My thought is just that it'll help us to reproduce SwiftUI issues as they come in and give us a better view into how Drop-in works (or doesn't) with SwiftUI.

If we want, we could even add a disclaimer to the README about how we don't officially support SwiftUI at this point. That way, we can be clear that the existence of this demo doesn't actually mean official SwiftUI support.

 ### Checklist

 - ~Added a changelog entry~

- @sestevens

### DIFF
--- a/Demo.xcodeproj/project.pbxproj
+++ b/Demo.xcodeproj/project.pbxproj
@@ -8,6 +8,8 @@
 
 /* Begin PBXBuildFile section */
 		4221CD56260E66D30025F6AD /* DemoPurchaseButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4221CD55260E66D30025F6AD /* DemoPurchaseButton.swift */; };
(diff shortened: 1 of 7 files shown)


sarahkoop (approved):
This looks good to me! I would include a note in the README about not officially supporting SwiftUI

sshropshire (approved):

sestevens:
> This looks good to me! I would include a note in the README about not officially supporting SwiftUI

Done!

scannillo (approved):
👏

PR #328, opened 2021-05-07
Fix bug with canceling out of PayPal flow

sestevens:
### Summary of changes

 - Fix a bug where we returned an error to the merchant if the user canceled out of the PayPal flow, instead of remaining on the payment selection sheet. 

**Context**: In v5, the Braintree SDK returns an error with the code `BTPayPalDriverErrorTypeCanceled` if the user cancels the PayPal flow. The PR updates Drop-in to check for that error code and do nothing if it's returned, so that the user can select a different payment method.

 ### Checklist

 - [x] Added a changelog entry

- @sestevens

### DIFF
--- a/Demo/Application/DemoDropInViewController.swift
+++ b/Demo/Application/DemoDropInViewController.swift
@@ -78,6 +78,7 @@ class DemoDropInViewController: DemoBaseViewController, DemoDropInViewDelegate {
             guard let result = result, error == nil else {
                 self.progressBlock?("Error: \(error!.localizedDescription)")
                 print("Error: \(error!)")
+                dropInController.dismiss(animated: true, completion: nil)
                 return
             }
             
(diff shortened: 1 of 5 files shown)


Demo/Application/DemoDropInViewController.swift
```diff
@@ -78,6 +78,7 @@ class DemoDropInViewController: DemoBaseViewController, DemoDropInViewDelegate {
             guard let result = result, error == nil else {
                 self.progressBlock?("Error: \(error!.localizedDescription)")
                 print("Error: \(error!)")
+                dropInController.dismiss(animated: true, completion: nil)
```
sestevens, line 81: Our UI tests would have caught this issue if we were dismissing the Drop-in when an error was returned back to the merchant.

sarahkoop (approved):

sshropshire (approved):

scannillo:
Did we check to see how the Venmo cancellation is currently working to see if we are doing the same thing?

sestevens:
> Did we check to see how the Venmo cancellation is currently working to see if we are doing the same thing?

Confirmed that the Venmo flow is working as expected!

PR #330, opened 2021-05-24
Remove ACKNOWLEDGEMENTS.md

gesa:
### Summary of changes

 - RIP

 ### Checklist

 - [x] Added a changelog entry

- @gesa


scannillo (approved):

sestevens:
Whoa, it looks like there are a lot of things in here we no longer use:
- NSURL+QueryDictionary, FLEX and iOS-Slide-Menu were removed [here](https://github.com/braintree/braintree-ios-drop-in/commit/fabb01557d552a0bb129cbfcbdf52e25f432bdc4)
- AFNetworking was removed [here](https://github.com/braintree/braintree-ios-drop-in/commit/979ae28bc31d1c754b01c155e405aa87510819fb)
- PureLayout was removed [here](https://github.com/braintree/braintree-ios-drop-in/commit/016671e46d0d65729713be5647ffac5bb2f2e128)
- I don't know when we stopped using Hockey, but we don't use it anymore

So we can remove everything except InAppSettingsKit and Braintree (although it's kind of weird that we're acknowledging ourselves 🤣 ).

sestevens:
I went ahead and removed the other things we don't use. I also removed Braintree since we refer to "3rd party libraries" at the top of the file, and Braintree isn't a 3rd party library for us.

sestevens:
We decided to remove the ACKNOWLEDGEMENTS.md file altogether since the only 3rd party framework being used in InAppSettingsKit, and it's only used in the demo app. The settings page of the demo app acknowledges that it was created with InAppSettingsKit.

PR #333, opened 2021-06-18
Increase Discover Card Range to 19 Digits

sarahkoop:
### Summary of changes

 - Increase valid Discover card length to 19 digits
 - Note: I couldn't find a valid Discover test card for that is greater than 16 digits, so was not able to end-to-end test this

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/BraintreeDropIn.podspec
+++ b/BraintreeDropIn.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
                        Check out our development portal at https://developers.braintreepayments.com.
   DESC
   s.homepage         = "https://www.braintreepayments.com/how-braintree-works"
-  s.documentation_url = "https://developers.braintreepayments.com/ios/start/hello-client"
+  s.documentation_url = "https://developer.paypal.com/braintree/docs/start/hello-client/ios/v5"
   s.screenshots      = ["https://github.com/braintree/braintree-ios-drop-in/raw/master/Images/client-sdk-ios-series-light.png", "https://github.com/braintree/braintree-ios-drop-in/raw/master/Images/client-sdk-ios-series-dark.png"]
   s.license          = "MIT"
   s.author           = { "Braintree" => "code@getbraintree.com" }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
 ## unreleased
-
 * Increase valid Discover card length to 19 digits
 
 ## 9.0.2 (2021-05-26)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+
+* Increase valid Discover card length to 19 digits
+
 ## 9.0.2 (2021-05-26)
 * Require `braintree_ios` v5.3.2 or higher
 * PayPal
--- a/Sources/BraintreeDropIn/Models/BTUIKCardType.m
+++ b/Sources/BraintreeDropIn/Models/BTUIKCardType.m
@@ -162,7 +162,7 @@
                                                         securityCodeName:BTDropInLocalizedString(CVC_FIELD_PLACEHOLDER)
                                                                 prefixes:@[@"^(5[1-5]|222[1-9]|22[3-9]|2[3-6]|27[0-1]|2720)\\d*"]];
         BTUIKCardType *discover = [[BTUIKCardType alloc] initWithBrand:BTDropInLocalizedString(CARD_TYPE_DISCOVER)
-                                                        securityCodeName:BTDropInLocalizedString(CVV_FIELD_PLACEHOLDER)
+                                                        securityCodeName:BTDropInLocalizedString(CID_FIELD_PLACEHOLDER)
                                                                 prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]
                                                          relaxedPrefixes:nil
                                                       validNumberLengths:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(16, 4)]
--- a/Sources/BraintreeDropIn/Models/BTUIKCardType.m
+++ b/Sources/BraintreeDropIn/Models/BTUIKCardType.m
@@ -162,8 +162,12 @@
                                                         securityCodeName:BTDropInLocalizedString(CVC_FIELD_PLACEHOLDER)
                                                                 prefixes:@[@"^(5[1-5]|222[1-9]|22[3-9]|2[3-6]|27[0-1]|2720)\\d*"]];
         BTUIKCardType *discover = [[BTUIKCardType alloc] initWithBrand:BTDropInLocalizedString(CARD_TYPE_DISCOVER)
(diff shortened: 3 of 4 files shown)


sshropshire (approved):
🌊

BraintreeDropIn.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
                        Check out our development portal at https://developers.braintreepayments.com.
   DESC
   s.homepage         = "https://www.braintreepayments.com/how-braintree-works"
-  s.documentation_url = "https://developers.braintreepayments.com/ios/start/hello-client"
+  s.documentation_url = "https://developer.paypal.com/braintree/docs/start/hello-client/ios/v5"
```
sarahkoop, line 13: This CocoaPods Build was failing in CI due to this link being broken. Updating this link resolved the CI failure.

scannillo: Random:

Looks like if we exclude the version #, the link doesn't work (ex: https://developer.paypal.com/braintree/docs/start/hello-client/ios/). This used to work, but something probably broke w/ the docs migration.

(https://developer.paypal.com/braintree/docs/start/hello-client) <-- this works though 🤷

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+
```
scannillo, line 4:
```suggestion
```

scannillo:
Can we get a sample test card from the Cards team?

BraintreeDropIn.podspec
sarahkoop: Yeah that's odd - I'll ask docs about that

Sources/BraintreeDropIn/Models/BTUIKCardType.m
```diff
@@ -162,8 +162,12 @@ + (NSArray *)allCards
                                                         securityCodeName:BTDropInLocalizedString(CVC_FIELD_PLACEHOLDER)
                                                                 prefixes:@[@"^(5[1-5]|222[1-9]|22[3-9]|2[3-6]|27[0-1]|2720)\\d*"]];
         BTUIKCardType *discover = [[BTUIKCardType alloc] initWithBrand:BTDropInLocalizedString(CARD_TYPE_DISCOVER)
-                                                      securityCodeName:BTDropInLocalizedString(CID_FIELD_PLACEHOLDER)
-                                                              prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]];
+                                                        securityCodeName:BTDropInLocalizedString(CVV_FIELD_PLACEHOLDER)
```
sestevens, line 165: Was this change from `CID_FIELD_PLACEHOLDER` to `CVV_FIELD_PLACEHOLDER` intentional? I think that would be a breaking change for merchants supplying their own localized strings files.

Sources/BraintreeDropIn/Models/BTUIKCardType.m
```diff
@@ -162,8 +162,12 @@ + (NSArray *)allCards
                                                         securityCodeName:BTDropInLocalizedString(CVC_FIELD_PLACEHOLDER)
                                                                 prefixes:@[@"^(5[1-5]|222[1-9]|22[3-9]|2[3-6]|27[0-1]|2720)\\d*"]];
         BTUIKCardType *discover = [[BTUIKCardType alloc] initWithBrand:BTDropInLocalizedString(CARD_TYPE_DISCOVER)
-                                                      securityCodeName:BTDropInLocalizedString(CID_FIELD_PLACEHOLDER)
-                                                              prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]];
+                                                        securityCodeName:BTDropInLocalizedString(CID_FIELD_PLACEHOLDER)
+                                                                prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]
+                                                         relaxedPrefixes:nil
+                                                      validNumberLengths:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(16, 4)]
```
sestevens: How come this is 16 instead of 19?

Sources/BraintreeDropIn/Models/BTUIKCardType.m
sarahkoop: Good catch that should not have changed - reverted it

Sources/BraintreeDropIn/Models/BTUIKCardType.m
```diff
@@ -163,7 +163,11 @@ + (NSArray *)allCards
                                                                 prefixes:@[@"^(5[1-5]|222[1-9]|22[3-9]|2[3-6]|27[0-1]|2720)\\d*"]];
         BTUIKCardType *discover = [[BTUIKCardType alloc] initWithBrand:BTDropInLocalizedString(CARD_TYPE_DISCOVER)
                                                       securityCodeName:BTDropInLocalizedString(CID_FIELD_PLACEHOLDER)
-                                                              prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]];
+                                                              prefixes:@[@"^(6011|65|64[4-9]|622)\\d*"]
+                                                       relaxedPrefixes:nil
+                                                    validNumberLengths:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(16, 4)]
```
sestevens, line 168: How come this is 16 instead of 19?

sarahkoop: I believe this makes a range starting from 16 with length 4 (so 16-19)

sestevens: Oh, I see! That makes sense.

sestevens (approved):

sarahkoop:
> Can we get a sample test card from the Cards team?

I got a test card to try out and it works to display the Discover icon when entering the card number, but it shows as Maestro after tokenizing in Sandbox... I think that's an issue with the test card not the SDK though.

scannillo:
> I got a test card to try out and it works to display the Discover icon when entering the card number, but it shows as Maestro after tokenizing in Sandbox... I think that's an issue with the test card not the SDK though.

The [`BTDropInResult` has a `paymentIcon` ](https://github.com/braintree/braintree-ios-drop-in/blob/c12e61046c63833a3ff90bc48b29c2f4edb54f3a/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h#L33)attribute which is what the Demo app uses to display the card icon (in this case looks like it is showing up as Maestro). I think we'll need to check this out and make sure we're returning the right card icon for these longer Discover cards.

sarahkoop:
> The [`BTDropInResult` has a `paymentIcon` ](https://github.com/braintree/braintree-ios-drop-in/blob/c12e61046c63833a3ff90bc48b29c2f4edb54f3a/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h#L33)attribute which is what the Demo app uses to display the card icon (in this case looks like it is showing up as Maestro). I think we'll need to check this out and make sure we're returning the right card icon for these longer Discover cards.

Yeah it looks like that icon is being set based on the Gateway response payment method type. I checked again and now it's returning Discover (and updating the icon appropriately). I think something was corrected on the back end, but I've reached out to confirm.

PR #338, opened 2021-06-29
Remove unused CardinalMobile for previous testing

scannillo:
### Summary of changes

- Before Cardinal released their xcframework, we had to have merchants manually include the `CardinalMobile.framework` into their apps in order to use SPM. This is no longer needed, since we can properly distribute the CardinalMobile.xcframework as a Swift Package.
- So ... this `CardinalMobile.framework` copy that lived in the `SampleApps/SPM` directory is no longer used/ needed for testing.

 ### Checklist

 - ~Added a changelog entry~

### Authors
@scannillo


sshropshire (approved):
🌊

PR #339, opened 2021-07-08
Collect Device Data

scannillo:
### Changes
- Expose `deviceData` on `BTDropInResult`

### Motivation
 - Android & Web DropIn offer device data collection capabilities.
 - In Android, you have to set `collectDeviceData` to true on your `DropInRequest` [here](https://github.com/braintree/braintree-android-drop-in/blob/481ef414dfde294c647534e0d570d3a175b5c770/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java#L87-L97). If you do, only then will you get back device data on your `DropInResult` [here](https://github.com/braintree/braintree-android-drop-in/blob/481ef414dfde294c647534e0d570d3a175b5c770/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInResult.java#L109-L116).
    - For iOS, I decided not to require merchants set the `collectDeviceData` boolean property on their request, and instead always include `deviceData` on the DropInResult.
    - I think since this is an additive (non-breaking) change, this is an easier integration pattern for merchants. They don't have to worry about requesting the data. They will always get it and can decide to ignore or use it.

### Concerns
I had a concern about performance, but used XCTest's `measure` function to time this test. I get roughly the same time to initialize the DropInResult with and without the call to PPDataCollector. I think it is OK to add.

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeDropIn/Models/BTDropInResult.m
+++ b/Sources/BraintreeDropIn/Models/BTDropInResult.m
@@ -11,26 +11,6 @@
 #import <BraintreeCore/BraintreeCore.h>
 #endif
 
-// Import PayPalDataCollector (Swift) module
-#if __has_include(<Braintree/Braintree-Swift.h>)      // CocoaPods
-#import <Braintree/Braintree-Swift.h>
-
-#elif SWIFT_PACKAGE                                   // SPM
-/* Use @import for SPM support
- * See https://forums.swift.org/t/using-a-swift-package-in-a-mixed-swift-and-objective-c-project/27348
- */
-@import PayPalDataCollector;
-
-#elif __has_include("Braintree-Swift.h")              // CocoaPods for ReactNative
-/* Use quoted style when importing Swift headers for ReactNative support
- * See https://github.com/braintree/braintree_ios/issues/671
- */
-#import "Braintree-Swift.h"
-
-#else                                                 // Carthage
-#import <PayPalDataCollector/PayPalDataCollector-Swift.h>
-#endif
-
 NSString *const BTDropInResultErrorDomain = @"com.braintreepayments.BTDropInResultErrorDomain";
 
 @implementation BTDropInResult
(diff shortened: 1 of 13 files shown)


Sources/BraintreeDropIn/Models/BTDropInResult.m
```diff
@@ -11,10 +11,43 @@
...
+#import <PayPalDataCollector/PayPalDataCollector-Swift.h>
+#endif
+
 NSString *const BTDropInResultErrorDomain = @"com.braintreepayments.BTDropInResultErrorDomain";
 
 @implementation BTDropInResult
 
+// For testing
+static Class PayPalDataCollectorClass;
+static NSString *PayPalDataCollectorClassString = @"PPDataCollector";
```
scannillo, line 40: I think we can do this with OCMock instead (see [docs](https://ocmock.org/reference/) under the 3.2 Stubs and verification section).

This project doesn't use OCMock, so I tried to import it. OCMock has an SPM bug, which I will report to their GitHub. So rather than use a 2nd package manager for the project to fetch OCMock, I went for this approach.

We can always change this testing situation in the future. Let me know what ya'll think.

sarahkoop: This approach seems reasonable to me. No need to introduce another package manager for testing.

sarahkoop (approved):
This looks good to me! Should we mark `collectDeviceData` on `BTDropInRequest` as deprecated if it's always going to be requested and exposed now?

scannillo:
> Should we mark collectDeviceData on BTDropInRequest as deprecated if it's always going to be requested and exposed now?

@sarahkoop There is no `collectDeviceData` on [BTDropInRequest](https://github.com/braintree/braintree-ios-drop-in/blob/master/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInRequest.h), so we should be good!

sshropshire (approved):

PR #342, opened 2021-07-13
Update Gems

sarahkoop:
### Summary of changes

- Address PR #341
- Run `bundle update`: Update `CFPropertySupport`, `activesupport`, `addressable`, `cocoapods`, `concurrent-ruby`, `ffi`, `i18n`, `jazzy`, `json`, `manitest`, `rake`, `rexml`, `rouge`, `tzinfo`, `xcodeproj`

 ### Checklist

### Authors

- @sarahkoop


Gemfile.lock
```diff
@@ -117,4 +119,4 @@ DEPENDENCIES
   xcpretty
 
 BUNDLED WITH
-   2.1.4
+   2.2.23
```
sarahkoop, line 122: Bumping the bundler version to the latest and to align with braintree_ios

scannillo (approved):

Gemfile.lock
scannillo: Also a note, users will need to `gem install bundler:2.2.23` to get the new bundler version. I think it's pretty self explanatory though so we don't need to worry about documenting it.

PR #352, opened 2021-09-07
Fix "current" section of reference docs

scannillo:
### Summary

Issue: The reference docs with "current" in their path for https://braintree.github.io/braintree-ios-drop-in/current/ was no accurately pointing to the latest version of the SDK. The reason it is now, is because I tested the following changes in this PR manually, via the Rakefile. Now the "current" section looks good.

### Changes

The changes really just update our jazzy workflow to use source-kitten, which the `braintree_ios` SDK was already doing. I honestly just copy pasted, then modified. At our next release, we can confirm that the `release.yml` changes work as expected.

- Updates Rakefile
- Update release.yml
- Move umbrella header to `Docs/` dir to mirror `braintree_ios` process

 ### Checklist

 - ~Added a changelog entry~
 - JIRA #DTBTSDK-1210

### Authors
@scannillo

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,7 +42,7 @@ class << self
   end
 
   def current_version
-    File.read(PODSPEC)[SEMVER]
+    "9.2.0"
   end
 
   def current_version_with_sha
--- a/Rakefile
+++ b/Rakefile
@@ -240,7 +240,7 @@ namespace :docs do
     run! "rm -rf docs_output"
     run(sourcekitten_objc_command)
     run(jazzy_command)
-    run! "rm swiftDoc.json && rm objcDoc.json"
+    run! "rm objcDoc.json"
     puts "Generated HTML documentation"
   end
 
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -205,14 +205,21 @@ end
 
 def jazzy_command
   %W[jazzy
-      --objc
+      --sourcekitten-sourcefile swiftDoc.json,objcDoc.json
```
scannillo, line 208:
```suggestion
      --sourcekitten-sourcefile objcDoc.json
```

sshropshire (approved):
🌊

scannillo:
`testTokenizeVenmo_whenAuthorizationSuccessful_returnsNonce`

This test is consistently failing. I think it has 0 to do with this PR and has to do with a new XCode 13 beta that was released.

scannillo:
Confirmed with manual testing that the latest iOS 15 beta is super wonky and causing that Venmo test to fail. We can ignore for this PR.

PR #366, opened 2022-02-07
Update Vault Manager Documentation

jaxdesmarais:
### Summary of changes

 - Update docs string to better reflect the behavior of how payment methods are returned from calling `mostRecentPaymentMethodForClientToken` function

 ### Checklist

 - [x] Added a changelog entry

- @Epreuve @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+* Update Vault Manager inline documentation 
+
 ## 9.4.0 (2022-01-18)
 * Require `braintree_ios` 5.6.1 or higher
 * Add specific error message for duplicate payment method
--- a/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h
+++ b/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h
@@ -56,7 +56,7 @@ typedef NS_ENUM(NSInteger, BTDropInErrorType) {
 /**
  * Fetch a `BTDropInResult` with the customer's vaulted payment method.
  * The payment method returned is not guaranteed to be the most recently added payment method.
- * If the last payment method selected from Drop-in was Apple Pay, a `BTDropInResult` with
+ * If the vaulted payment method is Apple Pay, a `BTDropInResult` with
  * `paymentMethodType == .applePay` will be returned in the completion block.
  *
  * @param clientToken Client token. Must be generated with a customer ID.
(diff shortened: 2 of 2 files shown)


sshropshire (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Braintree iOS Drop-in SDK - Release Notes
 
+## unreleased
+* Update Vault Manager inline documentation
```
sarahkoop, line 4: I'm not sure we need a CHANGELOG entry for this - does anyone feel strongly one way or the other?

jaxdesmarais: @Epreuve and I debated that but figured we could always remove it if anyone felt strongly one was or the other. I personally don't feel strongly about it either way.

sarahkoop: On second thought - I think it makes sense to keep it if we want to draw merchant attention to the fact that the behavior of this method is different than previously documented 👍

Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h
```diff
@@ -54,7 +54,7 @@ typedef NS_ENUM(NSInteger, BTDropInErrorType) {
 @property (nonatomic, strong, nullable) NSString *deviceData;
 
 /**
- * Fetch a `BTDropInResult` with the customer's most recently vaulted payment method.
+ * Fetch a `BTDropInResult` with the customer's vaulted payment methods. The order is not guaranteed.
```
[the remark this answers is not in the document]
jaxdesmarais: Ah you're right, good catch, we fetch the entire array but only return the first object in the array, which makes this docs string even more confusing since we cannot actually regularly determine what payment method that will be. I'll update this string.

Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h
```diff
@@ -54,13 +54,15 @@ typedef NS_ENUM(NSInteger, BTDropInErrorType) {
 @property (nonatomic, strong, nullable) NSString *deviceData;
 
 /**
- * Fetch a `BTDropInResult` with the customer's most recently vaulted payment method.
+ * Fetch a `BTDropInResult` with the customer's vaulted payment method.
+ * The payment method returned is not guaranteed to be the most recently added payment method.
  * If the last payment method selected from Drop-in was Apple Pay, a `BTDropInResult` with
```
sarahkoop, line 59:
```suggestion
 * If the vaulted payment method is Apple Pay, a `BTDropInResult` with
```
Nitpick - but this is probably also not necessarily the last payment method selected right?

sarahkoop (approved):

PR #372, opened 2022-03-28
Update braintree_ios Version to 5.8

kaylagalway:
### Summary of changes

This updates the braintree_ios framework version to 5.8

 ### Checklist

 - [x] Added a changelog entry

### Authors
- @jaxdesmarais @kaylagalway

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/braintree/braintree_ios",
         "state": {
           "branch": null,
-          "revision": "9a5da35d658c8c9f224424deaecae82c43bb3857",
-          "version": "5.7.0"
+          "revision": "914f9c60232d599ed61d52be1ece34b7e700faf6",
+          "version": "5.8.0"
         }
       }
     ]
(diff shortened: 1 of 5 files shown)


sshropshire (approved):
🌊

Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/braintree/braintree_ios",
         "state": {
           "branch": null,
-          "revision": "6be5b6ceabe6acf3353a986da272bc604c8cd010",
-          "version": "5.4.2"
+          "revision": "914f9c60232d599ed61d52be1ece34b7e700faf6",
+          "version": "5.8.0"
```
sarahkoop, line 10: Weird that this was 5.4.2 before

sarahkoop (approved):

Package.resolved
kaylagalway: Yeah, it actually didn't update/resolve correctly when we made the version change either. We had to try some manual commands to get it to the right version. May want to keep an eye on this in the future.

PR #378, opened 2022-08-09
Resolve `deviceData` nil on `BTDropInResult`

jaxdesmarais:
### Summary of changes

 - Resolve issue where `deviceData` was always nil on `BTDropInResult` - this was due to how we were mocking the data collector but previously never actually calling the data collector to get device data as expected

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais


sshropshire (approved):

Sources/BraintreeDropIn/Models/BTDropInResult.m
```diff
@@ -42,7 +42,13 @@ @implementation BTDropInResult
...
     self = [super init];
     if (self) {
-        _deviceData = [PayPalDataCollectorClass collectPayPalDeviceData];
+        // If we are testing we want to set PayPalDataCollectorClass to the class passed in `setPayPalDataCollectorClass`
+        if (PayPalDataCollectorClass != NSClassFromString(PayPalDataCollectorClassString)) {
+            _deviceData = [PayPalDataCollectorClass collectPayPalDeviceData];
+            return self;
+        }
+        // Otherwise we should use `PPDataCollector` to collect device data
+        _deviceData = [PPDataCollector collectPayPalDeviceData];
```
sshropshire, line 51: This is wild. So the test was injecting a definition for `PayPalDataCollectorClass`? So glad we're moving to Swift lol.

jaxdesmarais: Yeah, the test was setting a class, but the actual implementation never was - so tests were passing with the mock class being injected and the live code was returning nil for the past year 😢 Swift will certainly help us catch these cases sooner!

jaxdesmarais:
Tower was acting wild and closed/deleted this branch temporarily, so ignore that noise/chaos

mattwylder (approved):
🚢

PR #381, opened 2022-09-07
Require `braintree_ios` v5.12.0

scannillo:
### Summary of changes

 - Require `braintree_ios` [v5.12.0](https://github.com/braintree/braintree_ios/releases/tag/5.12.0) in podspec & package.swift
 - Enables DropIn to offer iOS 16 & Xcode 14 support

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## unreleased
 * Remove use of deprecated `setNetworkActivityIndicatorVisible` on iOS 13+ (the network activity indicator was removed from the status bar in iOS 13) (fixes #379)
-* Require `braintree_ios` 5.12.0 or higher
-  * Adds support for iOS 16 and Xcode 14
+* Add support for iOS 16 and Xcode 14
+  * Require `braintree_ios` 5.12.0 or higher
 
 ## 9.6.1 (2022-08-10)
 * Fix bug where `deviceData` on `BTDropInResult` was always returned as `nil`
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -14,7 +14,7 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(name: "Braintree", url: "https://github.com/braintree/braintree_ios", from: "5.9.0")
+        .package(name: "Braintree", url: "https://github.com/braintree/braintree_ios", from: "5.12.0")
```
jaxdesmarais, line 17: I think we normally run `swift package resolve` to update the package.resolved looking at previous PRs.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## unreleased
 * Remove use of deprecated `setNetworkActivityIndicatorVisible` on iOS 13+ (the network activity indicator was removed from the status bar in iOS 13) (fixes #379)
+* Require `braintree_ios` 5.12.0 or higher
+  * Adds support for iOS 16 and Xcode 14
```
jaxdesmarais, line 6: Take it or leave it, but should we move this point for adding support for iOS 16 + Xcode 14 to the main level vs nested? Wondering if this reads like we are just adding that support for core but not the drop-in?

scannillo: Good idea! Fixed

Package.swift
scannillo: Good catch! I forgot that our Demo app pulls in BT via SPM (hence tracked in the Package.resolved).

scannillo:
>⛵ it looks like the UI test that failed last is testDropIn_paypal_showAmount_receivesNonce. Looks like it's passing locally so we may just need to kick it again?

Kicked & passed!

sshropshire (approved):

Issue #383, opened 2022-11-30
mostRecentPaymentMethod callback fields are all null when fetching Apple Pay Card


jbloodwhatnot:
**Integration Details (please complete the following information):**
 - SDK/Library Version: 9.3.0 (Updated to latest version to e.g 9.7.0 and issue still happening_
 - Environment: Production (Issue not happening in Sandbox)
 - iOS Version: Seeing logs on both iOS 15 / 16
 - Xcode Version: 14.1
 - Device [e.g. iPhone 12]
 - Integration type & version Swift Package Manage
 - Development Processor Type: Intel, M1

**Describe the bug**
We are getting reports of users who have added apple pay cards that they are getting an error when paying for goods. After investigating the number of user reports we have verified that the users custom vault correctly has the apply pay cards added to their customer account. But the error is happening when we attempt to fetch the users most recent payment method. The response from the Braintree sdk callback has the error and result field as both as null. Which is causing the app to fail. Any idea what the issue could possibly be in order for both of these fields to return null.

Code snippet below:
```
BTDropInResult.mostRecentPaymentMethod(forClientToken: token) { res, error in
                guard error == nil else {
                    AppLogger.log(.error, message: "BRAINTREE FETCH ERROR", error: error)
                    continuation.resume(throwing: error!)
                    return
                }
                guard let res = res else {
                    AppLogger.log(.error, message: "BRAINTREE ERROR", error: PaymentError.noData)
                    continuation.resume(throwing: PaymentError.noData)
                    return
                }
                
                continuation.resume(returning: res)
            }
```

What we are seeing is the callback is returning both res and error field as nil. (Which I would have assumed one of these fields would always return non nil ? Any idea of what the cause for this could be.

From our logs it seems like its users using Apply Pay cards that are effected. It has started happening for our users for us since 31st October. I was able to reproduce locally but its hard to debug the issue as the SDK is not really giving any hints as to what has gone wrong here. I checked the users who had this issue and all had the Apple Cards correctly added to their paypal vaults/custom accounts.

**To Reproduce**
3. See error

**Expected behavior**
Description of what you expected to happen.

mostRecentPaymentMethod should return the users Apple Pay card.

**Screenshots**
N/A

jaxdesmarais:
Hello @jbloodwhatnot -

[...]

PR #387, opened 2023-01-04
Fix UnionPay Deprecation Issue

scannillo:
### Summary of changes

Addresses GH Issue --> https://github.com/braintree/braintree-ios-drop-in/issues/385
- Silence UnionPay method deprecation warnings
- Add `NEXT_MAJOR_VERSION` notes to remove UnionPay SMS flow

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo

PR #388, opened 2023-01-04
Add missing error handling for No Recent Payments

scannillo:
### Background

This GitHub Issue (https://github.com/braintree/braintree-ios-drop-in/issues/383) surfaced a case where we are returning back `(nil, nil)` to the merchant when they call [the `mostRecentPaymentMethod()` function](https://github.com/braintree/braintree-ios-drop-in/blob/89154896389b7d63f06f4808b571821701e1a383/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h#L66-L67).

Apparently, [our tests were asserting/expecting to return `(nil, nil)`](https://github.com/braintree/braintree-ios-drop-in/blob/89154896389b7d63f06f4808b571821701e1a383/UnitTests/BTDropInResultTests.swift#L38-L39) when no recent payment methods were found). I think this is weird and confusing behavior.

This PR treats this `(nil, nil)` case as a **bug** and fixes it by returning a proper merchant-facing error in the case of an empty vault list returned from the Gateway.

### Summary

- Add `BTDropInErrorTypeNoRecentPaymentMethods` error code
- Return error when no payment method nonces are returned from gateway, but the API call 200s

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeDropIn/Models/BTDropInResult.m
+++ b/Sources/BraintreeDropIn/Models/BTDropInResult.m
@@ -133,10 +133,15 @@ static NSUserDefaults *_userDefaults = nil;
                 BTPaymentMethodNonce *paymentMethod = paymentMethodNonces.firstObject;
                 result.paymentMethodType = [BTUIKViewUtil paymentMethodTypeForPaymentInfoType:paymentMethod.type];
                 result.paymentMethod = paymentMethod;
+                completion(result, nil);
+                return;
+            } else {
+                NSError *error = [[NSError alloc] initWithDomain:BTDropInResultErrorDomain
+                                                            code:BTDropInErrorTypeUnknown
+                                                        userInfo:@{NSLocalizedDescriptionKey: @"No recent payment methods found."}];
+                completion(nil, error);
+                return;
             }
-            
-            completion(result, nil);
-            return;
         }];
     }];
 }
--- a/Sources/BraintreeDropIn/Models/BTDropInResult.m
+++ b/Sources/BraintreeDropIn/Models/BTDropInResult.m
@@ -127,19 +127,19 @@ static NSUserDefaults *_userDefaults = nil;
                 completion(nil, error);
                 return;
             }
-            
-            if (paymentMethodNonces.count > 0) {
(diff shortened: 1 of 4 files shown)


scannillo:
CI failures addressed in https://github.com/braintree/braintree-ios-drop-in/pull/387 & https://github.com/braintree/braintree-ios-drop-in/pull/387#issuecomment-1371520757.

scannillo:
Bitcode CI failure addressed [here](https://github.com/braintree/braintree-ios-drop-in/pull/390).

jaxdesmarais:
> Bitcode CI failure addressed [here](https://github.com/braintree/braintree-ios-drop-in/pull/390).

Just merged this into master!

Sources/BraintreeDropIn/Models/BTDropInResult.m
```diff
@@ -133,10 +133,15 @@ + (void)mostRecentPaymentMethodForAPIClient:(BTAPIClient * _Nullable)apiClient
                 BTPaymentMethodNonce *paymentMethod = paymentMethodNonces.firstObject;
                 result.paymentMethodType = [BTUIKViewUtil paymentMethodTypeForPaymentInfoType:paymentMethod.type];
                 result.paymentMethod = paymentMethod;
+                completion(result, nil);
+                return;
+            } else {
```
jaxdesmarais, line 138: nit: I wonder if this would be more readable if we flipped this logic and did the error first if the payment method count is equal to 0. Not a blocker at all though!

jaxdesmarais (approved):

Sources/BraintreeDropIn/Models/BTDropInResult.m
scannillo: Great feedback, will update!

scannillo: Pushed!

sshropshire (approved):

mattwylder (approved):

PR #406, opened 2023-04-13
Fix of bug preventing input of further numbers when first digit is '1' in credit card expiry month

RyanLinXiang:
### Summary of changes

in iOS the expiry month input field for credit card does not accept any other number when the number '1' is entered as first digit. This is due to the `formattedValue` function accepting only first digits bigger than 1. The consequence is that the months **1**1 (November) and **1**2 (December) currently cannot be typed into the field as expiry month.

 ### Checklist

 - [x] Added a changelog entry

### Authors
- Ryan Lin Xiang

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Braintree iOS Drop-in SDK - Release Notes
+## 9.8.3 (2023-04-13)
+* Fixes bug preventing input of digit 1 as first number for credit card expiry month
 
 ## 9.8.2 (2023-04-10)
 * Silence UnionPay related deprecation warnings introduced in `braintree_ios` 5.21.0 and higher.
--- a/Sources/BraintreeDropIn/Models/BTUIKCardExpiryFormat.m
+++ b/Sources/BraintreeDropIn/Models/BTUIKCardExpiryFormat.m
@@ -16,9 +16,17 @@
         return;
     }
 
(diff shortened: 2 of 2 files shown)


hollabaq86:
👋 @RyanLinXiang thanks for the PR! We'll take a look and provide feedback.

For internal tracking, issue 8321

scannillo:
👋 Hi @RyanLinXiang - thanks for making a PR to the Braintree Drop in SDK!

We are having trouble understanding the problem that your PR is trying to solve, as we don't see any issues when typing in expiration months that start with `1`. Can you provide more information, and maybe a video, describing your UI bug? Do you have steps to replicate this bug, as well?

RyanLinXiang:
> 👋 Hi @RyanLinXiang - thanks for making a PR to the Braintree Drop in SDK!
> 
> We are having trouble understanding the problem that your PR is trying to solve, as we don't see any issues when typing in expiration months that start with `1`. Can you provide more information, and maybe a video, describing your UI bug? Do you have steps to replicate this bug, as well?

Hi Scannillo - never mind. We are using the React Native version of the SDK (https://github.com/wgltony/react-native-braintree-dropin-ui) and the error happens there. Since it is based on your SDK. I thought the bug originates here. But after installing the demo app I just found out that the bug does not occur with your SDK. You can close the PR.

scannillo:
@RyanLinXiang thanks for following up! Closing.

PR #409, opened 2023-05-04
Drop Xcode 13 + require Xcode 14.1

jaxdesmarais:
### Summary of changes

 - On April 25, 2023 Apple is mandating that apps submitted to the App Store must be built with Xcode 14.1 or higher (Swift 5.7.1+)
    - [See news article](https://developer.apple.com/news/?id=jd9wcyov#:~:text=Starting%20April%2025%2C%202023%2C%20iOS,on%20the%20Mac%20App%20Store)
- Update all GitHub workflows with new Xcode requirement
- Remove Xcode 13 support from GitHub workflows
- Update Package.swift with new Swift requirement
- Update podspec with new Swift requirement
- Add README and CHANGELOG entry with this change

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ on: [pull_request, workflow_dispatch]
 jobs:
   cocoapods:
     name: CocoaPods (Xcode 14)
-    runs-on: macOS-latest
+    runs-on: macOS-12
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
@@ -13,7 +13,7 @@ jobs:
         run: pod lib lint
   spm:
     name: SPM (Xcode 14)
-    runs-on: macOS-latest
+    runs-on: macOS-12
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -3,46 +3,22 @@ on: [pull_request, workflow_dispatch]
 jobs:
   cocoapods:
     name: CocoaPods (Xcode 14)
-    runs-on: macOS-12
+    runs-on: macOS-latest
```
sshropshire, line 6: 🧶 : keeping the explicit version here could prevent the possibility of new macOS versions breaking the CI pipeline.

sshropshire (approved):
🌊

.github/workflows/build.yml
jaxdesmarais: It looks like we opted for latest in the core equivalent PR: https://github.com/braintree/braintree_ios/pull/1018.

scannillo: That's a good call - down to keep the explicit versions if we think it's more stable!

jaxdesmarais: Updated here: https://github.com/braintree/braintree-ios-drop-in/pull/409/commits/a299a36aca829ffb730156f608e9c42aa400d1af

jaxdesmarais:
Seems like UI tests have been flake central in general but especially in the Drop-in. Gonna ignore it for now and made internal ticket DTBTSDK-2695 for this work.

PR #410, opened 2023-05-08
Update UI Tests

jaxdesmarais:
### Summary of changes

 - Cleanup `XCTestCase` extensions
     - We were not using `file` or `line` so can simplify these methods quite a bit
     - Updated these methods to match the implementation used in `braintree_ios`
 - Bump timeouts to 20s
 - Remove 3DS v1 UITests
     - 3DSv1 is no longer supported so understandably 3DS v1 UITests were failing
     - 3DSv2 UITests exist in the `BraintreeDropIn_ThreeDSecure_2_UITests` class
 - Update missing postal code field on test

 ### Checklist

- @jaxdesmarais


Demo/UITests/BraintreeDropIn_UITests.swift
```diff
@@ -138,6 +138,11 @@ class BraintreeDropIn_securityCodeValidation_CardForm_UITests: XCTestCase {
         securityCodeField.forceTapElement()
         securityCodeField.typeText("200")
 
+        let postalCodeField = elementsQuery.textFields["12345"]
```
jaxdesmarais, line 141: The "Add Card" button is not hittable until all fields are filled so we needed to fill in the postal code element for this test

Demo/UITests/BraintreeDropIn_UITests.swift
```diff
@@ -700,151 +705,6 @@ class BraintreeDropIn_PayPal_Disabled_UITests: XCTestCase {
     }
 }
 
-class BraintreeDropIn_ThreeDSecure_UITests: XCTestCase {
```
jaxdesmarais, line 703: These are 3DSv1 tests - the v2 tests still exist below but we do not need these tests because the legacy flow is no longer supported by Cardinal

Demo/UITests/BraintreeDropIn_Localization_UITests.swift
```diff
@@ -18,7 +18,7 @@ class BraintreeDropIn_Localization_UITests: XCTestCase {
         app.buttons["Add Payment Method"].tap()
 
         expectation(for: NSPredicate(format: "exists == true"), evaluatedWith: app.staticTexts["Seleccionar forma de pago"])
-        waitForExpectations(timeout: 10.0)
+        waitForExpectations(timeout: 20)
```
scannillo, line 21: Dang, were these really timing out? 10s is a lot on it's own

Demo/UITests/Helpers.swift
```diff
@@ -1,57 +1,38 @@
 import XCTest
 
 extension XCTestCase {
-    func waitForElementToAppear(_ element: XCUIElement, timeout: TimeInterval = 10,  file: String = #file, line: Int = #line) {
+
+    func waitForElementToAppear(_ element: XCUIElement, timeout: TimeInterval = 20) {
```
scannillo, line 5: I hesitate slightly about the bump to 20s. It feels like if it takes over 20s to load, then our test are actually not good

scannillo: I would still love to move to snapshot tests, they're quick and reliable. But it seems pretty far off on our priority list right now.

Demo/UITests/BraintreeDropIn_Localization_UITests.swift
jaxdesmarais: Yeah some of them were legit timing out looking through the logs 😢

Demo/UITests/Helpers.swift
jaxdesmarais: We default to `15` on `braintree_ios` but I know even there sometimes our UITests time out. I think it has more to do with the runners than anything we do. If they take less than 20 seconds it won't wait around though so that's nice. More just giving the runners extra time if they needed it. I'm fine matching BT core if we want but again, we may have to kick them often like we do on BT core.

jaxdesmarais: Huge ➕ on moving to snapshot testing. I'm down to mess around with it one open dev if you're down to explore it more. I've not dealt with it much but down to explore!

scannillo: True good call - a lot of it must boil down to GH Actions resources. Down to mess with Snapshot testing sometime! I've worked with a bit prior. Another cool thing you can do is UI test specific components (such as individual buttons and subviews).

scannillo (approved):

Demo/UITests/BraintreeDropIn_Localization_UITests.swift
sshropshire: That is a lot for a UI test unless it's making networking calls or performing disk i/o.

sshropshire (approved):
🌊

PR #411, opened 2023-05-08
Update Docs Portion of GitHub Release

jaxdesmarais:
### Summary of changes

The docs job has been failing for a while and we've needed to manually complete this step each time we do a release. It seems like it was always attempting to pull the release tag before it was generated vs running the job after the other job completed. Now that we require Xcode 14.1 in our `release.yml` we can remove this as a separate job as it was only done because SourceKitten required a higher Xcode + Swift version.

 ### Checklist

- @jaxdesmarais


jaxdesmarais:
Fix for UITest failures is here: https://github.com/braintree/braintree-ios-drop-in/pull/410. We expect them to fail until that PR is merged in.

scannillo (approved):

PR #430, opened 2023-10-19
Restructure Demo - Use CocoaPods

scannillo:
### Summary

We want to start converting the DropIn source code file-by-file from Objective-C to Swift. Starting this work revealed a limitation of the structure of our Demo app and local xcodeproject.

See error:
Screenshot 2023-10-19 at 11 28 57 AM

Currently our Demo uses SPM to pull in a local version of the DropIn source code. It is a known limitation that SPM does not support mixed language dependencies. So, with this setup we will not be able to run our Demo app as we convert file-by-file to Swift.

### Changes

This PR restructures our Demo app to pull in a local version of DropIn via CocoaPods, instead of SPM.

- Added a `Podfile` for the Demo app
   - This means that our Demo app looks at our local [`BraintreeDropIn.podspec`](https://github.com/braintree/braintree-ios-drop-in/blob/main/BraintreeDropIn.podspec) in order to build the DropIn
- Removed SPM frameworks from Demo xcodeproj

 ### Checklist

 - ~Added a changelog entry~

### Authors
@scannillo


scannillo:
I am messing around with how development with this structure will feel, so far the ObjC <--> Swift interop is tricky. We might prefer a "Carthage" style setup for our demo instead.

scannillo:
Closing in favor of using a Carthage-style setup for our project structure

PR #431, opened 2023-10-23
Use hardcoded ClientToken in UITests

scannillo:
### Summary of changes

 - This PR updates our demo app settings to support using hard-coded/static ClientToken values versus fetching them in our UI tests
- The goal of this PR is to reduce UI test flakiness on CI by eliminating the client-token fetch network request
- Introduce `DemoAuthType` enum to mirror braintree_ios [test setup](https://github.com/braintree/braintree_ios/pull/1067)

 ### Checklist

 - ~Added a changelog entry~

### Authors
@jaxdesmarais @scannillo


scannillo:
> Failing after 46m

Gonna 🦵 kick the UI tests again just to see how long it takes

scannillo:
UI tests expected to fail

sarahkoop (approved):

jaxdesmarais:
It seems like the tests with the client tokens with customer IDs are failing with an "invalid auth fingerprint" error (such as the `BraintreeDropIn_EditMode_UITests` tests). I wonder if there is a different validation that happens when validating an auth fingerprint that includes a client ID? We may need to work with the authy team to get a long lived client token for UI tests that includes a customer ID or poke around at how Android/Web handles these UI tests. Alternatively we can go back to fetching a client token for the customer ID case, but I assume that'd still result in flaky tests.

scannillo:
> It seems like the tests with the client tokens with customer IDs are failing with an "invalid auth fingerprint" error (such as the BraintreeDropIn_EditMode_UITests tests). I wonder if there is a different validation that happens when validating an auth fingerprint that includes a client ID? We may need to work with the authy team to get a long lived client token for UI tests that includes a customer ID or poke around at how Android/Web handles these UI tests. Alternatively we can go back to fetching a client token for the customer ID case, but I assume that'd still result in flaky tests.

Great catch! I was also able to replicate that locally. So I removed the `DemoAuthType.uiTestHardcodedClientTokenWithCustomerID` option and replaced the tests that used it with regular `-ClientToken` setting. The tests are now passing locally!

jaxdesmarais (approved):
🔥

PR #448, opened 2024-04-11
Bump `braintree_ios` to  5.25.0

KunJeongPark:
### Summary of changes

 - Modify Package.swift and Podspec files to point to BT iOS 5.25.0
 - Update CHANGELOG

 ### Checklist

 - [x] Added a changelog entry

- @KunJeongPark @agedd

### DIFF
--- a/BraintreeDropIn.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/BraintreeDropIn.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,25 +1,23 @@
 {
-  "object": {
-    "pins": [
-      {
-        "package": "Braintree",
-        "repositoryURL": "https://github.com/braintree/braintree_ios",
-        "state": {
-          "branch": null,
-          "revision": "3924f5ea0295087081e7e637fbf1333361e7054b",
-          "version": "5.23.0"
-        }
-      },
-      {
-        "package": "InAppSettingsKit",
-        "repositoryURL": "https://github.com/futuretap/InAppSettingsKit.git",
-        "state": {
-          "branch": null,
-          "revision": "71e03ebad8462618f96c575c86d2e9220f880676",
-          "version": "3.3.2"
-        }
+  "pins" : [
+    {
+      "identity" : "braintree_ios",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/braintree/braintree_ios",
+      "state" : {
+        "revision" : "619e78a9ab9423be4871baf406546c22573086cf",
+        "version" : "5.25.0"
       }
-    ]
-  },
-  "version": 1
+    },
+    {
+      "identity" : "inappsettingskit",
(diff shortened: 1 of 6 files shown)


BraintreeDropIn.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,25 +1,23 @@
...
+      "identity" : "inappsettingskit",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/futuretap/InAppSettingsKit.git",
+      "state" : {
+        "revision" : "71e03ebad8462618f96c575c86d2e9220f880676",
+        "version" : "3.3.2"
+      }
+    }
+  ],
+  "version" : 2
```
jaxdesmarais, line 22: Can we do some digging and just make sure there aren't any issues moving this to version 2?

Package.resolved
```diff
@@ -1,16 +1,14 @@
...
+      "state" : {
+        "revision" : "619e78a9ab9423be4871baf406546c22573086cf",
+        "version" : "5.25.0"
       }
-    ]
-  },
-  "version": 1
+    }
+  ],
+  "version" : 2
```
jaxdesmarais, line 13: Same question here on just triple checking this won't causes issues moving from v1 to v2

BraintreeDropIn.xcworkspace/xcshareddata/swiftpm/Package.resolved
KunJeongPark: Yes, that's what I am looking up. They are also missing the "object" key that was there before. 
So I think something is wrong.

KunJeongPark: I found this post: https://forums.swift.org/t/using-swiftpm-5-6-and-supporting-older-versions/55708

KunJeongPark: It is recommended by a few people here is to add Package.resolved to .gitignore file

jaxdesmarais: We want to leave it so that we are all using the same version as far as adding it to the `.gitignore`. Lets run a test on 15.0 which is our minimally supported Xcode version in the SDK. If that looks good we can apply this change as is.

KunJeongPark: SPMTest looks good with Xcode 15.0 pointing to this branch.

jaxdesmarais: Thanks for triple checking! ❤️

jaxdesmarais (approved):

agedd (approved):

